PR #2, opened 2018-12-20
Radios

jasperswart:


packages/asc-core/src/Radios/Radios.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
```
chrisvanmook: Personally I'm not a big fan of a component that renders another component in this case (besides it's also not very react-ish, we would like to use the asc-core only for styled-components and not react). 
Perhaps create a (Radio)Bar / (Radio)Wrapper of some sort that only have styles for the wrapper itself, that can render children, but does not need to know anything about it's children 😃

packages/asc-core/src/Radio/Radio.tsx
```diff
@@ -60,7 +62,7 @@ const Wrapper = styled.div`
 `
 
 const InputRadio = (props: any) => (
```
chrisvanmook: Like discussed, we don't use react in the asc-core package 👿 😉 . Consider moving this to the asc-ui?

packages/asc-core/src/Radio/Radio.tsx
```diff
@@ -50,8 +50,10 @@ const Label = styled.label`
   }
 `
 
-const Wrapper = styled.div`
+const Wrapper = styled.span`
```
chrisvanmook: Nice to get started on creating components now!! 🎉

packages/asc-core/src/Radio/Radio.tsx
```diff
@@ -50,8 +50,10 @@ const Label = styled.label`
...
   position: relative;
+  display: ${props => props.orientation === 'horizontal' ? 'inline-block' : 'block' };
+  margin-right: ${props => props.orientation === 'horizontal' ? '24px' : '0' };
```
chrisvanmook: For the PoC we can use these arbitrary values like 24px, but eventually we might want to consider using styled-system or sth else. 😄 https://github.com/jxnblk/styled-system

PR #3, opened 2019-02-01
Radios again

jasperswart:


packages/asc-ui/src/Radio/Radio.tsx
```diff
@@ -1,8 +1,13 @@
 import * as React from 'react'
-import { Radio as RadioStyled } from '@amsterdam/asc-core'
-
+import { Radio as RadioStyled, RadioLabel, RadioItem } from '@amsterdam/asc-core'
+console.log(RadioItem);
```
chrisvanmook: hmmm 😛

packages/asc-ui/src/Radio/Radio.tsx
```diff
@@ -1,8 +1,13 @@
...
 const Radio = (props: any) => (
-  <RadioStyled {...props} />
+  <div>
```
chrisvanmook: Don't think you need to use div here, as RadioItem already has one.

packages/asc-ui/src/Radio/Radio.stories.tsx
```diff
@@ -1,11 +1,12 @@
...
+import { ButtonBar } from '@amsterdam/asc-core'
+import Radio from './Radio'
 
-storiesOf('Radio', module).add('All Radios', () => (
+storiesOf('Radio', module).add('All radios', () => (
   <div>
-    <Radio id="radio-1" name="gender" value="female" label="Vrouw" defaultChecked />
-    <Radio id="radio-2" name="gender" value="male" label="Man" />
-    <Radio id="radio-3" name="gender" value="none" label="Onzijdig" disabled />
+    <Radio id="female" name="gender" value="female" label="Vrouw" />
```
chrisvanmook: I think we can leave out the id, right? Since id is always the same as value, you can pass value to prop id in ./Radio

packages/asc-ui/src/Radio/Radio.tsx
```diff
@@ -1,8 +1,13 @@
 import * as React from 'react'
-import { Radio as RadioStyled } from '@amsterdam/asc-core'
-
+import { Radio as RadioStyled, RadioLabel, RadioItem } from '@amsterdam/asc-core'
+console.log(RadioItem);
 const Radio = (props: any) => (
```
chrisvanmook: I think we should use props here, then it would be clear for the user (IDE's etc.) which props he can pass

chrisvanmook: Yes exactly, but also state the prop types in typescript :)

packages/asc-ui/src/Radio/Radio.stories.tsx
chrisvanmook: sorry, I wasnt clear what I meant hehe, I think in this case the user (read: developer) shouldn't worry about any id, since we can handle the connection between for & id in the component. We just have to make sure we generate a unique id, in case someone uses the same value on the same page.

packages/asc-ui/src/Radio/Radio.tsx
jasperswart: weg nu

packages/asc-ui/src/Radio/Radio.tsx
jasperswart: removed!

packages/asc-ui/src/Radio/Radio.tsx
jasperswart: I refactored that

PR #6, opened 2019-03-26
Add socialbar and socialbutton component, extended from the button an…

royderks:
…d buttonbar component


packages/asc-ui/src/components/SocialBar/SocialBar.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { AscCore } from '@datapunt/asc-core'
+import { ReactComponent as Facebook } from '@datapunt/asc-assets/lib/Icons/Facebook.svg'
+import { ReactComponent as Twitter } from '@datapunt/asc-assets/lib/Icons/Twitter.svg'
+import { ReactComponent as Linkedin } from '@datapunt/asc-assets/lib/Icons/Linkedin.svg'
+import { ReactComponent as Email } from '@datapunt/asc-assets/lib/Icons/Email.svg'
+import { ReactComponent as Print } from '@datapunt/asc-assets/lib/Icons/Print.svg'
+import SocialButton from '../SocialButton'
+
+const SocialBar: React.FC<{}> = ({ children }) => (
+  <AscCore.SocialBar>
```
chrisvanmook: I don't think we should pre-define all the socialbuttons in this component, this is what the user should do

packages/asc-core/src/components/Button/Button.ts
```diff
@@ -79,4 +79,20 @@ export const IconButton = styled(ButtonBase)<Props>`
   padding: 5px;
 `
 
+export const SocialButton = styled(ButtonBase)<Props>`
+  padding: 0px;
+  width: 30px;
```
janjaap: Shouldn't this value be configurable as a component prop?
Also, the `ButtonBase` component doesn't have styling for `:focus`. Shouldn't the focus state styling be defined here for the `SocialButton` component?

packages/asc-core/src/components/Icon/Icon.ts
```diff
@@ -6,17 +6,20 @@ import getThemeColor from '../../utils/getThemeColor'
 type Props = {
   color?: Theme.TypeLevel
   iconUrl?: string
+  size?: number
```
janjaap: Why not give `size` a default value of `20`?

packages/asc-core/src/components/SocialBar/SocialBar.ts
```diff
@@ -0,0 +1,8 @@
+import styled from '../../styled-components'
+import ButtonBar from '../ButtonBar'
+
+type Props = {}
+
+const SocialBar = styled(ButtonBar)<Props>``
```
janjaap: If you don't pass in styling, is it still useful to make ButtonBar a styled component?

packages/asc-ui/src/components/SocialBar/SocialBar.tsx
janjaap: Well, it gives developers an idea of what they can use.

chrisvanmook: Then I think it should be moved to a story / documentation example rather than a standalone component

royderks: Moved implementation to a story example

packages/asc-core/src/components/Button/Button.ts
royderks: Added :focus. For now the social button just reuses the styling and should always be 30px. Can be changed later on with a default of 30 if needed

packages/asc-core/src/components/Icon/Icon.ts
royderks: 👍

packages/asc-core/src/components/SocialBar/SocialBar.ts
royderks: 👍

royderks:
Icons still up for check by design, functionality should be complete

package.json
```diff
@@ -34,62 +34,63 @@
...
-    "@types/enzyme": "^3.1.18",
+    "@types/enzyme": "^3.9.1",
     "@types/enzyme-adapter-react-16": "^1.0.5",
     "@types/jest": "^24.0.11",
-    "@types/node": "^11.11.1",
-    "@types/react": "^16.8.7",
-    "@types/react-dom": "^16.8.2",
+    "@types/node": "^11.12.0",
+    "@types/react": "^16.8.8",
+    "@types/react-dom": "^16.8.3",
```
janjaap: Shouldn't both `react` and `react-dom` have the same version?

royderks: This was added by merging with https://github.com/Amsterdam/amsterdam-styled-components/pull/7

janjaap: I understand, but shouldn't all versions of `react`, `react-dom`, `@types/react` and `@types/react-dom` be in sync?

royderks: Yes, true

PR #7, opened 2019-03-27
Added svgo, to use in building asc-assets

chrisvanmook:


packages/asc-assets/.svgo.yml
```diff
@@ -0,0 +1,3 @@
+plugins:
+  - removeViewBox: false
+  - removeDimensions: true
```
janjaap: There is a lot more optimisations we could potentially benefit from, like:
```yaml
- removeDoctype
- removeComments
- removeMetadata
- removeXMLNS
- removeEditorsNSData
- cleanupIDs
- removeRasterImages
- removeUselessDefs
- removeUnknownsAndDefaults
- removeUselessStrokeAndFill
- removeHiddenElems
- removeEmptyText
- removeEmptyAttrs
- removeEmptyContainers
- removeUnusedNS
- removeDesc
- prefixIds
- ...
```

packages/asc-assets/package.json
```diff
@@ -20,7 +20,7 @@
     "assets": "assets"
   },
   "scripts": {
-    "build": "mkdir -p lib && cp -R src/* lib/"
+    "build": "mkdir -p lib && cp -R src/* lib/ && svgo --config=.svgo.yml -f ./src/Icons/ -o ./lib/Icons"
```
janjaap: Why not use SVGO in a Webpack loader?

packages/asc-core/src/components/Icon/Icon.ts
```diff
@@ -15,8 +15,6 @@ const Icon = styled.span<Props>`
   height: 20px;
 
   & > svg {
-    width: 20px;
```
janjaap: Is this change related to adding SVGO?

packages/asc-assets/package.json
chrisvanmook: Because we don't create a bundle, we only provide optimized svg's in the package. It's up to the user of asc-assets (developer) how to bundle the code with the svg's.

packages/asc-core/src/components/Icon/Icon.ts
chrisvanmook: Yes and no. It's not directly linked to the svgo, however, I noticed while playing with svgo config, these lines could be removed as they don't do anything (because now we get rid of the width & height attributes)

janjaap: Nice

packages/asc-assets/.svgo.yml
chrisvanmook: I didn't really dove into this in detail yet. I only added the optimisations I knew we would need for now. But would be great if we could use these! Do you have any experience with the optimisations you listed above?

janjaap: Most of them, yes.

PR #14, opened 2019-04-08
Readme review

aditudorache:


README.md
```diff
@@ -29,7 +31,7 @@ The guidelines can be found here (TBD)
 
 ## Trade off
```
janjaap: Might as well change this to 'Trade-off' or 'Tradeoff'

README.md
```diff
@@ -51,21 +53,35 @@ The guidelines can be found here (TBD)
 
 ## Usage
 
-##### Development
+### Development
+
+```bash
```
janjaap: `bash` Isn't a supported syntax highlighter. Check https://github.com/github/linguist/blob/master/vendor/README.md for reference.

README.md
aditudorache: ok

README.md
aditudorache: This seem to work in github md. Please see:
https://stackoverflow.com/questions/20303826/highlight-bash-shell-code-in-markdown

README.md
aditudorache: ok

README.md
janjaap: `bash` Isn't recognised as a highlight identifier. If you remove it, you won't see a difference.

PR #17, opened 2019-04-09
Dp 6803 share bar

aditudorache:


packages/asc-ui/src/components/Modal/Modal.stories.tsx
```diff
@@ -55,7 +55,7 @@ storiesOf('Modal', module)
       onClose={action('close modal from escape button or click outside')}
       blurredNode={window.document.querySelector('#root') as HTMLInputElement}
     >
-      <TopBar>
+      <TopBar backgroundColor="lightgrey">
```
royderks: only use colors imported from the theme

packages/asc-ui/src/components/TopBar/StyledTopBar.ts
```diff
@@ -0,0 +1,28 @@
+import styled from 'styled-components'
+
+const defaultBackgroundColor = '#fff'
```
royderks: Use color from the theme

packages/asc-ui/src/components/TopBar/TopBar.stories.tsx
```diff
@@ -6,7 +6,7 @@ import Typography from '../Typography'
 import IconButton from '../IconButton'
 
 const SimpleComponent: React.FC<{}> = () => (
-  <TopBar>
+  <TopBar backgroundColor="grey">
```
royderks: Use color from the theme

packages/asc-ui/src/components/TopBar/StyledTopBar.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+const StyledTopBar = styled.header<Props>`
+  display: flex;
+  position: relative;
+  align-items: center;
+  z-index: 0;
+  min-height: 54px;
+  padding: 0 15px;
+  background-color: ${props =>
```
royderks: The props are destructured here in most of the codebase

packages/asc-ui/src/components/TopBar/StyledTopBar.ts
royderks: Preferably create an object containing the const defaultProps = { ... }

packages/asc-ui/src/components/Modal/Modal.stories.tsx
aditudorache: done

packages/asc-ui/src/components/TopBar/StyledTopBar.ts
aditudorache: done

packages/asc-ui/src/components/TopBar/StyledTopBar.ts
aditudorache: done

packages/asc-ui/src/components/ShareButton/StyledShareButton.ts
```diff
@@ -0,0 +1,20 @@
+import styled from 'styled-components'
```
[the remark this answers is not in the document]
aditudorache: Unfortunately, the way the styles are separated from the ui has no benefits while both libraries are depending on react. With the current structure is the development speed decreased because the dependent libraries must be build each tiem. asc-core would make sense when the css styling would be agnostic and would be used in an 'asc-ui' for React, for Vue, for Angular.

packages/asc-core/src/utils/README.md
```diff
@@ -0,0 +1,49 @@
+Styles utility functions
```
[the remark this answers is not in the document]
aditudorache: correct, done

packages/asc-ui/src/components/TopBar/TopBar.stories.tsx
aditudorache: done

packages/asc-ui/package.json
```diff
@@ -13,7 +13,9 @@
     "@datapunt/asc-assets": "^0.6.2-alpha.0",
     "@datapunt/asc-core": "^0.6.2-alpha.0",
     "@types/styled-components": "^4.1.1",
-    "classnames": "^2.2.6"
+    "classnames": "^2.2.6",
+    "jest-styled-components": "^6.3.1",
+    "styled-components": "4.1.3"
```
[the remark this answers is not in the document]
aditudorache: we need to do that for speeding up the development. will discuss a solution with the other frontenders

chrisvanmook:
I'm aware that core uses react under the hood for now and that this may
speed up the development process by adding styles in ui. However, I think
it's not the right place / moment to already make these decisions in this
branch. I think for now we do have to keep the two packages separated until
we find a solution for this problem. Otherwise it'll become messy and even
more unclear for other developers in how to deal with this. Let alone we
will start to loose our possible solution of making asc framework agnostic,
even if it isn't for now.

packages/asc-ui/src/components/ShareButton/ShareButton.tsx
```diff
@@ -1,25 +1,25 @@
 import React from 'react'
-import { AscCore } from '../../styles'
 import { Icon } from '../..'
+import ShareButtonStyle, {
+  Props as ShareButtonProps,
```
royderks: Move to export in ButtonStyle

packages/asc-ui/src/components/TextField/__snapshots__/TextField.test.tsx.snap
```diff
@@ -19,7 +19,11 @@ exports[`<TextField /> should render 1`] = `
   color: black;
 }
 
+<<<<<<< HEAD:packages/asc-ui/src/components/TextField/__snapshots__/TextField.test.tsx.snap
```
royderks: Check

packages/asc-ui/src/components/TopBar/TopBar.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react'
-import StyledTopBar, { Props } from '../../styles/components/StyledTopBar'
+import TopBarStyle, { Props } from '../../styles/components/TopBarStyle'
```
royderks: ->>  TopBarStyleProps

packages/asc-ui/src/components/ShareButton/ShareButton.tsx
aditudorache: done

packages/asc-ui/src/components/TextField/__snapshots__/TextField.test.tsx.snap
aditudorache: done

packages/asc-ui/src/components/TopBar/TopBar.tsx
aditudorache: done

PR #19, opened 2019-04-10
Added functions to easily get typography and colors from the theme

royderks:


packages/asc-core/src/utils/themeUtils.ts
```diff
@@ -0,0 +1,28 @@
+import { Theme } from '../theme'
+import { fromTheme } from '.'
+
+export function getTypographyFromTheme(
```
janjaap: Can't this be a curried function? Like so:
```js
export function getTypographyFromTheme(attributeType: string) {
    return function(theme: Theme.ThemeInterface) {
        return fromTheme(`typography.${[attributeType]}`)({ theme });
    }
}
```
Should work in styled component template literal by doing
```js
font-family: ${getTypographyFromTheme('fontFamily')};
```
instead of
```js
font-family: ${({ theme }) => getTypographyFromTheme(theme, 'fontFamily')};

```

packages/asc-core/src/utils/themeUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+  theme: Theme.ThemeInterface,
+  attributeType: string,
+) {
+  return fromTheme(`typography.${[attributeType]}`)({ theme })
+}
+
+export function getColorFromTheme(
```
janjaap: Same here

packages/asc-core/src/utils/themeUtils.ts
royderks: Nice addition, added issue to do this later https://github.com/Amsterdam/amsterdam-styled-components/issues/22

PR #24, opened 2019-04-14
DP-6803 Buttons styles refactored

aditudorache:


packages/asc-core/src/components/ButtonBaseStyle/index.ts
```diff
@@ -0,0 +1 @@
+export { default, Props } from './ButtonBaseStyle'
```
royderks: ButtonBaseStyleProps

packages/asc-core/src/components/shared/focus.ts
```diff
@@ -2,10 +2,11 @@ import { css } from '../../styled-components'
 import { Theme } from '../../theme'
 import { getColorFromTheme } from '../../utils'
 
-export default (theme: Theme.ThemeInterface) => css`
+const focus = (theme: Theme.ThemeInterface) => css`
```
royderks: rename getFocusStyle

royderks: To themeUtils

PR #33, opened 2019-04-23
Dp 6846 header tall

aditudorache:

- [x] The component styles are only placed in the `asc-core/src/components` folder


packages/asc-ui/src/components/Button/Button.test.tsx
```diff
@@ -1,20 +1,11 @@
...
-      <ThemeProvider>
-        <ButtonBar>
-          <Button color="primary">Hello</Button>
-          <Button color="secondary">World</Button>
-        </ButtonBar>
-      </ThemeProvider>,
-    )
+describe('Button', () => {
+  it('should render the button', () => {
+    const component = shallow(<Button color="primary">Hello</Button>).dive()
```
royderks: Do you need the `dive()` to test this component, have a look at the snapshot

packages/asc-ui/src/components/Header/HeaderTallTitle.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+const HeaderTallTitleStyle = styled.div`
+  & > h1 {
+    display: flex;
+    height: 100%;
+    margin: 0;
+  }
+
+  & > h1 > span {
```
royderks: Why not extend the styling for `HeaderTitleStyle`. Writing it like this takes out the advantage of `styled-components` and we could just use inline css instead

packages/asc-ui/src/components/Icon/Icon.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+import Icon from './Icon'
+import 'jest-styled-components'
+
+describe('Icon', () => {
+  it('should render the icon', () => {
+    const component = shallow(
+      <Icon inline size={24} padding={4}>
+        icon-content
+      </Icon>,
+    ).dive()
```
royderks: See if we want to use `dive()` in instances like this

packages/asc-ui/src/components/Header/HeaderTallTitle.tsx
aditudorache: I don't think I understand what you mean. Can you post an example?

packages/asc-ui/src/components/Header/Header.stories.tsx
```diff
@@ -2,76 +2,75 @@ import * as React from 'react'
...
-          homeLink="http://data.amsterdam.nl"
-          search={<CenteredElement> Search </CenteredElement>}
-          menu={<CenteredElement> M </CenteredElement>}
-          headerSize="short"
-        />
-        <ContentFiller backgroundColor={contentBackgrountColor} />
-      </BoxWrapper>
-    </>
-  </ThemeProvider>
+  <BoxWrapper backgroundColor={outsideBackgoundColor}>
```
aditudorache: why is the theme provider removed?

packages/asc-core/src/components/HeaderStyle/headerSizes.ts
```diff
@@ -0,0 +1,3 @@
+export const HEADER_SIZES = {
+  CONTENT_HEIGHT_TALL: 60,
```
aditudorache: is this not 50?

packages/asc-ui/src/components/Header/Header.stories.tsx
chrisvanmook: Because we already use it via .storybook/config.js, so we won't need to include them all the time :)

packages/asc-ui/src/components/AmsterdamLogo/AmsterdamLogo.tsx
```diff
@@ -0,0 +1,18 @@
...
+import AmsterdamLogoStyle from '../../styles/components/AmsterdamLogo/AmsterdamLogoStyle'
+
+type Props = {
+  tall?: boolean
+  href?: string
+}
+
+const AmsterdamLogo: React.FC<Props> = ({ tall, ...otherProps }) => (
+  <AmsterdamLogoStyle {...{ tall }} {...otherProps}>
+    <LogoShort />
```
aditudorache: this construction gives the impression that both logo's are to be renderd here. Could be this be rendered depending on the tall property?

packages/asc-ui/src/components/Header/Header.stories.tsx
aditudorache: NICE!

packages/asc-ui/src/components/ContextMenu/__tests__/__snapshots__/ContextMenu.test.tsx.snap
```diff
@@ -6,7 +6,7 @@ exports[`ContextMenu should render 1`] = `
   onKeyDown={[Function]}
 >
   <ContextMenuButton
-    arrowIcon={<ReactComponent />}
+    arrowIcon={<UNDEFINED />}
```
aditudorache: i think this is something wrong but is out of scope of this pr

PR #39, opened 2019-05-06
Theme selectors improvements

aditudorache:

- [x] The component styles are only placed in the `asc-core/src/components` folder. Exception is the `asc-ui/src/internals` folder where the styles are allowed. The components from this folder are just for internal use in the stories.

### DIFF
--- a/packages/asc-core/src/components/ButtonStyle/ButtonStyle.ts
+++ b/packages/asc-core/src/components/ButtonStyle/ButtonStyle.ts
@@ -1,6 +1,6 @@
 import { readableColor } from 'polished'
 import styled from '../../styled-components'
-import { fillSvgFromTheme, getColorFromTheme } from '../../utils'
+import { svgFill, getColorFromTheme } from '../../utils'
 import ButtonBaseStyle, { ButtonBaseStyleProps } from '../ButtonBaseStyle'
 
 export type Props = ButtonBaseStyleProps
@@ -25,7 +25,7 @@ const ButtonStyle = styled(ButtonBaseStyle)<Props>`
   & svg {
     width: 30px;
     height: 30px;
-    ${({ color, theme }) => fillSvgFromTheme(theme, color)};
+    ${({ color, theme }) => svgFill(color)({ theme })};
   }
 `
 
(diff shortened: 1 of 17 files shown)


packages/asc-core/src/utils/themeUtils.ts
```diff
@@ -20,39 +29,42 @@ export const getTypographyFromTheme = (
...
   })
-  return breakpoint && breakpoint(type)
+  return breakpointFunc && breakpointFunc(type)
 }
 
-export const fillSvgFromTheme = (
-  theme: Theme.ThemeInterface,
+export const svgFill = (
   colorType?: Theme.TypeLevel,
   variant: string = 'main',
```
chrisvanmook: we can type the variants, as we know which ones are restricted (cannot be only just any string). perhaps we can use keyof here?

packages/asc-core/src/utils/themeUtils.ts
```diff
@@ -3,6 +3,15 @@ import { css } from '../styled-components'
 import { Theme } from '../theme'
 import { fromTheme } from '.'
 
+export const color = (
+  colorType?: Theme.TypeLevel,
+  variant: string = 'main',
```
chrisvanmook: we can type the variants, as we know which ones are restricted (cannot be only just any string). perhaps we can use keyof here?

packages/asc-core/src/components/ButtonStyle/ButtonStyle.ts
```diff
@@ -25,7 +25,7 @@ const ButtonStyle = styled(ButtonBaseStyle)<Props>`
   & svg {
     width: 30px;
     height: 30px;
-    ${({ color, theme }) => fillSvgFromTheme(theme, color)};
+    ${({ color, theme }) => svgFill(color)({ theme })};
```
chrisvanmook: plz find a clever way to curry theme to svgFill

packages/asc-core/src/utils/themeUtils.ts
aditudorache: will be done in the second iteration

packages/asc-core/src/utils/themeUtils.ts
aditudorache: will be done in the second iteration

packages/asc-core/src/components/ButtonStyle/ButtonStyle.ts
aditudorache: will be done in the second iteration

PR #64, opened 2019-07-01
(Chore) Grid components

janjaap:


packages/asc-core/src/components/RowStyle/RowStyle.ts
```diff
@@ -0,0 +1,144 @@
...
+} from '../../utils/grid'
+
+type FlexJustify =
+  | 'flex-start'
+  | 'flex-end'
+  | 'center'
+  | 'space-between'
+  | 'space-around'
+  | 'space-evenly'
+type FlexAlign = 'flex-start' | 'center' | 'flex-end' | 'stretch' | 'baseline'
```
chrisvanmook: perhaps you could put these Flex types to a generic file? Since you've declared the same type in ColumnStyle.ts

packages/asc-ui/src/components/Grid/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as Column } from './Column'
```
chrisvanmook: Can we also export these in the main index.ts file?

PR #65, opened 2019-07-02
Chore column row components

janjaap:
This PR:
- Extends the theme configuration with layouts; a description of how elements on a page can be laid out. 
The configuration takes the following shape:
  ```javascript
  {
    layout1: {
      columns: 24,
      margin: 40,
      gutter: 20,
      min: 1024,
      max: 1080
    },
    layout2: {
      columns: 16,
      margin: 30,
      gutter: 12,
      min: 960,
      max: 1024
    },
    ...
  }
  ```
  The props `margin`, `gutter` and `columns` are required. `min` And `max` don't have to be present, but at least one of the needs to be configured to be able to construct the grid.

- Adds components `Row` and `Column`. The former ensures that its children are positionable in a, well, row. It does so by applying `flex-box` style rules. By default, a `Row`'s children are spaced with space between them. A `Row` component takes a `debug` prop. Setting this prop will render a label that shows the name of the layout as well as the min and max values of that layout. It will also render a linearly repeated background image that shows the position of the columns in the grid. This helps with laying out elements:

Screenshot 2019-07-02 at 21 36 00

  __Spanning__
  The latter (`Column`) can be configured to span a certain amount of columns at all layouts or an amount of columns per layout:
  ```jsx
  // always span two columns, regardless of layout
  <Column span={2}>...</Column>

  // conditionally span columns
  <Column span={{ layout1: 4, layout2: 3 }}>...</Column>
  ```

  A column can also be pushed a number of columns to the right as can it be given an order the same way it can be given a span:

  __Ordering and pushing__
  ```jsx
  // ordering columns
  <Column span={{ layout1: 1, layout2: 2 }}>...</Column>
  <Column span={{ layout1: 2, layout2: 1 }}>...</Column>

  // pushing a column
  <Column push={2}>...</Column>
  <Column push={{ layout1: 0, layout2: 1 }}>...</Column>
  ```

  __Debugging__
  A `Column` component can also be passed a `debug` prop. Setting this prop to 'true' will render a label that shows the current span of that column. As the layout changes, the label on the column will also change value.

  __Nesting__
  Nesting columns can be tricky, because of the width of the parent element. The `Column` element contains its own parent span context and can thus safely be nested:

  ```jsx
  <Column span={10}>
    <Column span={4}>...</Column>
    <Column span={6}>...</Column>
  </Column>
  ```
_(Screenshot showing two columns (`center #1` and `center #2`) nested inside the middle column that spans 8 columns in the 'big' layout configuration)_

  Spanning, ordering and pushing can be applied to nested columns as well.

- Adds utility functions to aid grid construction and column width calculation


packages/asc-ui/src/components/Grid/Column.tsx
```diff
@@ -0,0 +1,42 @@
...
+  parentSpan?: Theme.TypeSpan
+}
+
+const ParentContext = React.createContext<Context>({})
+
+const Column: React.FC<Props> = ({ children, span, id, ...props }) => {
+  const { parentSpan } = React.useContext<Context>(ParentContext)
+
+  return (
+    <ParentContext.Provider value={{ parentSpan: span }}>
```
royderks: When is this context provider used?

packages/asc-ui/src/components/Grid/Column.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+type Props = {
+  children: React.ReactNode
+  debug?: boolean
+  debugColor?: string
+  id?: keyof Theme.GridInterface
+  order?: Theme.TypeSpan
+  push?: Theme.TypeSpan
+  span: Theme.TypeSpan
+  styles?: any
```
royderks: Rename to css to match other components?

packages/asc-ui/src/components/Grid/Column.tsx
janjaap: Context is necessary to allow columns to be nested. All column widths are relative and having a relative width of an element that's inside another element that already has a relative width will set a wrong value for the width unless the nested element is aware of the width of its parent.

packages/asc-ui/src/components/Grid/Column.tsx
janjaap: Hm, maybe I should remove that.

packages/asc-ui/src/components/Grid/Column.tsx
royderks: Ah. Couldn't find any other Consumers components or useContext hooks in the PR

janjaap: Not even on line 23? ;)

royderks: Ah yes, but won't parentSpan always be the same as the span prop in this scenario? Or am I missing some context?

janjaap: Well, the point is that `parentSpan` (which comes from the context) will have the value of `span` which comes from the component that wraps another component.

PR #67, opened 2019-07-05
(Chore) grid tests

janjaap:
This PR:
- applies the correct theme configuration
- introduces `notes` add-on for Storybook
- introduces `@testing-library/react` package
- makes defining layouts more generic so that the `layouts` config can contain any number of layout configuration with any name
- adds test for both `Row` and `Column` grid components as well as for the grid utility functions
- adds documentation and examples for the grid components
- fixes an issue in grid utilities where columns were placement was calculated incorrectly
- ensures that styled components use the correct theme


janjaap:
@chrisvanmook Comments addressed. I also pushed a couple of other fixes.

janjaap:
@chrisvanmook I changes the grid max width value

PR #70, opened 2019-07-12
Added hygen + generator templates :tada:

chrisvanmook:
Hygen is a code generator https://www.hygen.io/


janjaap:
I created a component `YoMama` with `hygen` and gave `YoGrandma` as the parent, but this generates a `YoMama` component in the `YoMama` folder which is generated in the `YoGrandma` folder. Is that supposed to happen?

_templates/component/new/index.ejs.t
```diff
@@ -0,0 +1,12 @@
...
+to: packages/asc-ui/src/components/<%= parent %>/<%=name%>/index.ts
+---
+import <%=name%>Style from './<%=name%>Style'
+export { default as <%=name%> } from './<%=name%>'
+
+const <%=name%>Styles = {
+  <%=name%>Style,
+}
+export {
+  <%=name%>Styles
```
janjaap: Other components export their styled counterpart as `...Style` instead of `...Styles`. Change of approach?

_templates/component/new/story.ejs.t
```diff
@@ -0,0 +1,13 @@
+---
+to: "<%= story ? `packages/asc-ui/src/components/${parent}/${name}/${name}.story.tsx` : null %>"
+skip_if: <%= story %>
+---
+
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import <%= name %> from './<%= name %>'
+
+storiesOf('Atom/<%= name %>', module)
```
janjaap: Shouldn't `Atom` also be configurable?

_templates/component/new/index.ejs.t
chrisvanmook: Mistake, I'll remove this :)

_templates/component/new/prompt.js
```diff
@@ -13,11 +13,17 @@ module.exports = [
   {
     type: 'input',
     name: 'parent',
-    message: "What is the component's parent? (Leave empty if not applicable)",
+    message:
+      "Does the component have an existing parent? If so, what's it's name? (Leave empty if not applicable)",
```
janjaap: Should be "what's its name" instead of "what's it's name"

royderks: Yes

PR #80, opened 2019-07-23
Added ImageCard components, moved 'strong' prop to typography

chrisvanmook:

- [x] The component styles are only placed in the `asc-core/src/components` folder. Exception is the `asc-ui/src/internals` folder where the styles are allowed. The components from this folder are just for internal use in the stories.

### DIFF
--- a/packages/asc-core/src/theme/default/typography.ts
+++ b/packages/asc-core/src/theme/default/typography.ts
@@ -26,7 +26,7 @@ const typography: Theme.TypographyInterface = {
     fontWeight: 700,
     letterSpacing: 'inherit',
     lineHeight: '28px',
-    marginBottom: '0',
+    marginBottom: '8px',
     breakpoints: {
       tabletS: {
         fontSize: '24px',
--- a/packages/asc-ui/src/components/ImageCard/ImageCard.stories.tsx
+++ b/packages/asc-ui/src/components/ImageCard/ImageCard.stories.tsx
@@ -11,7 +11,7 @@ import { breakpoint } from '../../utils'
 import { ImageCardWrapperStyle } from './ImageCardStyle'
 import TypographyStyle from '../Typography/TypographyStyle'
 
-const ImageCardConstainer = styled.section`
+const ImageCardContainer = styled.section`
   width: calc(100% + 24px);
   margin-left: -12px;
   margin-right: -12px;
@@ -88,7 +88,7 @@ storiesOf('Atoms/ImageCard', module)
           <Heading as="h6" styleAs="h2">
             Jeugdwerkloosheid Amsterdam daalt naar 6,2%
           </Heading>
-          <Paragraph skipAutoMargin>
+          <Paragraph gutterBottom={0}>
             Amsterdamse jongeren even vaak werkloos als gemiddeld in Nederland
           </Paragraph>
         </ImageCardContent>
@@ -99,7 +99,7 @@ storiesOf('Atoms/ImageCard', module)
     <Row>
       <Column wrap span={{ small: 1, medium: 2, big: 6, large: 8, xLarge: 8 }}>
         <div style={{ width: '100%' }}>
-          <ImageCardConstainer>
+          <ImageCardContainer>
             <ImageCardWrapperBig>
               <ImageCard
                 margin={12}
@@ -109,7 +109,7 @@ storiesOf('Atoms/ImageCard', module)
                   <Heading as="h6" styleAs="h2">
                     Jeugdwerkloosheid Amsterdam daalt naar 6,2%
                   </Heading>
-                  <Paragraph skipAutoMargin>
+                  <Paragraph gutterBottom={0}>
                     Amsterdamse jongeren even vaak werkloos als gemiddeld in
                     Nederland
                   </Paragraph>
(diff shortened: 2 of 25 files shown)


packages/asc-core/src/theme/default/typography.ts
```diff
@@ -26,7 +26,7 @@ const typography: Theme.TypographyInterface = {
     fontWeight: 700,
     letterSpacing: 'inherit',
     lineHeight: '28px',
-    marginBottom: '0',
+    marginBottom: '8px',
```
aditudorache: have you checked the side effects of this change?

packages/asc-ui/src/components/ImageCard/ImageCard.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { ImageCardContent } from './ImageCardContent'
+import Heading from '../Heading'
+import Paragraph from '../Paragraph'
+import Row from '../Grid/Row'
+import Column from '../Grid/Column'
+import { breakpoint } from '../../utils'
+import { ImageCardWrapperStyle } from './ImageCardStyle'
+import TypographyStyle from '../Typography/TypographyStyle'
+
+const ImageCardConstainer = styled.section`
```
aditudorache: Maybe Container or Constrainer?

packages/asc-ui/src/components/ImageCard/ImageCard.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+    width: calc(100% + 4px);
+  }
+`
+const ImageCardWrapperBig = styled.div`
+  flex-basis: 100%;
+  @media screen and ${breakpoint('min-width', 'tabletM')} {
+    flex-basis: ${100 - 100 / 3}%;
+  }
+`
+const ImageCardWrapperSmall = styled.div`
```
aditudorache: these styles could be moved to separate files or, if that specific, maybe to atlas

packages/asc-ui/src/components/ImageCard/ImageCardContent/ImageCardContentStyle.ts
```diff
@@ -0,0 +1,18 @@
+import styled from '@datapunt/asc-core'
+import { color, breakpoint } from '../../../utils'
+
+export type Props = {}
```
aditudorache: remove if no properties needed

packages/asc-ui/src/components/ImageCard/ImageCardStyle.ts
```diff
@@ -0,0 +1,67 @@
...
+  backgroundImage: string
+  loading?: boolean
+}
+
+export type WrapperProps = {
+  margin?: number
+}
+
+export const ImageCardWrapperStyle = styled.div<WrapperProps>`
+  ${({ margin }) =>
```
aditudorache: why not 
margin: ${({margin}) => margin? `${margin}px` : '0'};?

packages/asc-ui/src/components/Typography/TypographyStyle.ts
```diff
@@ -1,13 +1,18 @@
 import styled, { css, styledComponents, Theme } from '@datapunt/asc-core'
 import { em, margin } from 'polished'
-import { getTypographyFromTheme, color as getColor } from '../../utils'
+import { getTypographyFromTheme, color as getColor, color } from '../../utils'
```
aditudorache: color and getColor are the same. Maybe rename the color util function to something like getColor to prevent the naming conflict with the color css prop?

packages/asc-ui/src/utils/themeUtils.ts
```diff
@@ -41,23 +41,32 @@ export const breakpoint = (
...
+const generateCSSFromTypography = (
+  {
+    defaultColor,
+    fontWeight,
+    fontSize,
+    letterSpacing,
+    lineHeight,
+    marginBottom,
+  }: any,
+  skipAutoMargin?: boolean,
```
aditudorache: this naming is not the best choice because is specific to margin-bottom. Maybe something like noMarginBottom? The construction of the function is very specific, don't have any sugesstiion now but please review this, maybe it can be done in a different way

packages/asc-ui/src/components/ImageCard/ImageCard.stories.tsx
chrisvanmook: why? these are just used for the atlas implementation  story

packages/asc-ui/src/utils/themeUtils.ts
chrisvanmook: good one! I will instead pass the existing gutterBottom instead, since we already have it. we can later rename this one as well, but lets make sure it's backwards compatible

packages/asc-core/src/theme/default/typography.ts
chrisvanmook: doesn't affect anything in ASC, will add this to the changelist

packages/asc-ui/src/components/ImageCard/ImageCard.stories.tsx
aditudorache: ok, so thet are to be implemented in atlas as well. nice!

PR #84, opened 2019-07-29
Also have the option to neglect the max width from the theme and have…

royderks:
… the grid displayed full screen

### DIFF
--- a/packages/asc-ui/src/components/Grid/RowStyle.ts
+++ b/packages/asc-ui/src/components/Grid/RowStyle.ts
@@ -26,6 +26,7 @@ export type TypeProps = {
   debugColor?: string
   halign?: FlexJustify
   hasMargin?: boolean
+  hasMaxWidth?: boolean
   valign?: TypeFlexPosition
 }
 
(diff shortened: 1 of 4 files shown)


packages/asc-ui/src/components/Grid/RowStyle.ts
```diff
@@ -35,7 +36,8 @@ const RowStyle = styled.div<TypeProps>`
   display: flex;
   justify-content: ${({ halign }: { halign?: string }) => halign};
   align-items: ${({ valign }: { valign?: string }) => valign};
-  max-width: ${fromTheme('maxGridWidth')}px;
+  max-width: ${({ hasMaxWidth, theme }) =>
```
janjaap: Did you also consider overriding the theme instead of having an extra prop?

royderks: We have some pages that will use the grid and in theory could increase in width indefinitely depending on the screen size or zoom level. Setting this in the theme to 9999999 might be an option otherwise?

janjaap: It might, but I'm not sure how easy the implementation of that use case is. The proposed solution might indeed be better.

PR #90, opened 2019-07-31
Created tag component, with backgroundcolor from the theme and readab…

royderks:
…le color using polished


packages/asc-ui/src/components/Tag/TagStyle.ts
```diff
@@ -0,0 +1,30 @@
...
+import Typography from '../Typography'
+import { color as getColor } from '../../utils'
+
+export type Props = {
+  backgroundColor?: keyof Theme.Tint | keyof Theme.TypeLevel
+} & TypographyProps &
+  styledComponents.StyledProps<any>
+
+export const TagStyleCSS = ({ backgroundColor, theme }: any) => {
+  const color: string = backgroundColor.startsWith('level')
```
royderks: Might move this to utils

package.json
```diff
@@ -23,6 +23,7 @@
     "build:watch": "yarn build:ts -w",
     "clean:generated": "find ./packages -name \"*.d.ts\" -type f -delete && find ./packages -name \"*.tsbuildinfo\" -type f -delete",
     "deploy-storybook": "storybook-to-ghpages",
+    "generate:component": "hygen component new",
```
chrisvanmook: The problem with this is that we cannot use commands like `hygen component new:stories` (this will only add a stories file)...

packages/asc-ui/src/components/Tag/TagStyle.ts
```diff
@@ -0,0 +1,30 @@
+import styled, { css, styledComponents, Theme } from '@datapunt/asc-core'
+import { readableColor } from 'polished'
+import { Props as TypographyProps } from '../Typography/TypographyStyle'
+import Typography from '../Typography'
+import { color as getColor } from '../../utils'
+
+export type Props = {
+  backgroundColor?: keyof Theme.Tint | keyof Theme.TypeLevel
```
chrisvanmook: backgroundColor indicates that you can just set a color code. Perhaps change it to variant? and have a enum set of variants that will link to a limited set of colors?

chrisvanmook: btw, we also have to do this with our Button component...

packages/asc-ui/src/components/Tag/TagStyle.ts
royderks: Moved to utils

package.json
royderks: Deleted

packages/asc-ui/src/components/Tag/TagStyle.ts
royderks: Changed to `color` for now to even it with the `Button`. However `colorType` might be a better fitl, but results in breaking changes...

PR #153, opened 2019-09-11
Add a negative tabIndex to the div element wrapping the Toggle compon…

royderks:
…ent, this will make sure the element is focusable and therefore can be blurred

### DIFF
--- a/packages/asc-ui/src/components/Toggle/Toggle.tsx
+++ b/packages/asc-ui/src/components/Toggle/Toggle.tsx
@@ -104,6 +104,7 @@ const Toggle: React.FC<Props> = ({
       onBlur={handleOnBlur}
       onKeyDown={handleOnKeyDown}
       {...otherProps}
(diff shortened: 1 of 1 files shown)


packages/asc-ui/src/components/Toggle/Toggle.tsx
```diff
@@ -104,6 +104,7 @@ const Toggle: React.FC<Props> = ({
       onBlur={handleOnBlur}
       onKeyDown={handleOnKeyDown}
       {...otherProps}
+      tabIndex={-1} // This will enable the onblur event for this div element on all browsers
```
chrisvanmook: no side effects? also in searchbartoggle?

royderks: Will also work over there, the negative tabindex makes it possible to focus and blur programmatically but not by keyboard 

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

PR #179, opened 2019-10-22
Add flex item alignment

timvanoostrom:
This will align the Chevron correctly with the text.


packages/asc-ui/src/components/Link/LinkStyle.ts
```diff
@@ -40,6 +40,7 @@ export const InlineLinkStyleCSS = css`
 `
 export const DefaultLinkStyleCSS = css<Props>`
   display: inline-flex;
+  align-items: center;
```
chrisvanmook: Perhaps add a case on line 78? `case 'with-chevron': css'align-items: center;'` with this style added, so it won't effect other links

timvanoostrom:
Will fix later

PR #182, opened 2019-10-29
Add an example of linking a package in readme documentation. Update t…

ebabel-eu:
Update the styling of the Tag component so that its background colour does not touch the edge of text below it when it is part of a line of text that wraps, like on the homepage.

Add an example of linking a package in readme.


packages/asc-ui/src/components/Tag/TagStyle.ts
```diff
@@ -20,6 +20,7 @@ export const TagStyleCSS = ({ colorType, colorSubtype, theme }: Props) => {
     width: fit-content;
     background-color: ${backgroundColor};
     color: ${readableColor(backgroundColor)};
+    line-height: 2em;
```
royderks: How does this center and align when it's using in example in a `h1` element?

PR #187, opened 2019-10-29
Radio button

robinpiets:
To do:
- [x] add `variants` with styling
- [x] add `name` prop to RadioGroup to pass to its children
- [x] fix radio without radiogroup
- [x] add styling (https://designsystem.amsterdam.nl/7awj1hc9f/p/93c181)
- [x] add error state
- [x] add radio label positions (top, right, bottom, left)
- [x] add `active` context from Label
- [x] add horizontal/vertical prop and style
- [x] write proper test for click()
- [x] export/expose RadioGroup
- [x] update changelog


packages/asc-ui/src/components/Radio/Radio.tsx
```diff
@@ -1,27 +1,85 @@
...
-  <RadioItemStyle>
-    <RadioLabelStyle htmlFor={id} disabled={disabled}>
-      {label || value}
-    </RadioLabelStyle>
-  </RadioItemStyle>
-)
+  // Make the label aware of changes in the checked state
+  useMemo(() => {
+    setActive(checked)
+  }, [checked, setActive])
```
janjaap: Should `setActive` be a dependency here? It's a function and should not change, should it?

packages/asc-ui/src/components/Radio/Radio.tsx
```diff
@@ -1,27 +1,85 @@
...
+      <RadioStyle
+        {...{
+          ...otherProps,
+          disabled,
+          id,
+          defaultChecked,
+          name,
+        }}
+        onFocus={() => {
+          setFocus(true)
```
janjaap: Can't this be solved by CSS alone? Using `:focus` style rule, for instance.

packages/asc-ui/src/components/Radio/Radio.tsx
```diff
@@ -1,27 +1,85 @@
...
-}
+const Radio: React.FC<Props & React.HTMLAttributes<HTMLDivElement>> = ({
+  className,
+  onChange,
+  variant,
+  disabled,
+  id,
+  defaultChecked,
+  error: errorProp,
+  ...otherProps
```
janjaap: Is `value` one of the `otherProps`? If so, wouldn't a test that verifies that a radio button has the correct value be in order?

packages/asc-ui/src/components/Radio/RadioStyle.ts
```diff
@@ -1,29 +1,139 @@
...
+  border: 1px solid;
+  border-radius: 50%;
+  ${({ checked }) =>
+    checked &&
+    css`
+      color: ${getVariantColor()};
+      &::after {
+        content: '';
+        position: absolute;
+        height: 16px;
```
aditudorache: themeSpacing?

packages/asc-ui/src/components/Radio/RadioStyle.ts
```diff
@@ -1,29 +1,139 @@
...
+        width: 16px;
```
aditudorache: themeSpacing

packages/asc-ui/src/components/Radio/Radio.tsx
robinpiets: I need `focus` as a prop:
```
${({ checked, disabled, focus }) =>
    !checked &&
    !disabled && // IE11 fix. If an element is disabled, you can still hover on it on IE11.
    !focus && // don't override the outline if element is also focussed
    css`
```

packages/asc-ui/src/components/Radio/Radio.tsx
robinpiets: Good point.`value` is one of the `otherProps` indeed.

packages/asc-ui/src/components/Radio/RadioStyle.ts
robinpiets: No `themeSpacing` on `height` and `width`.

packages/asc-ui/src/components/Radio/RadioStyle.ts
robinpiets: No `themeSpacing` on `height` and `width`.

packages/asc-ui/src/components/Radio/Radio.tsx
robinpiets: @chrisvanmook has a reason for this

packages/asc-ui/src/components/Radio/Radio.tsx
robinpiets: Added `value` to props and rewritten the test

packages/asc-ui/src/components/Radio/Radio.tsx
janjaap: Checked this; has to do with the linter rules and the React core team recommendations on how dependencies should be used.

packages/asc-ui/src/components/Radio/RadioStyle.ts
```diff
@@ -1,29 +1,142 @@
...
+export { RadioCircleStyle }
+
+const RadioWrapperStyle = styled.div<StyleOnlyProps & { focus: boolean }>`
+  position: relative;
+  display: inline-flex;
+  user-select: none;
+  vertical-align: middle;
+  flex-shrink: 0; /* IE11 fix */
+  padding: ${themeSpacing(1)};
+  margin-bottom: 1px;
```
chrisvanmook: why the one pixel?

robinpiets: image
Because the spacing is 1px in design system (see screenshot)

PR #190, opened 2019-10-30
Text Button variant update: no padding and auto height

robinpiets:

### DIFF
--- a/packages/asc-ui/src/components/Button/ButtonStyle.ts
+++ b/packages/asc-ui/src/components/Button/ButtonStyle.ts
@@ -46,6 +46,13 @@ export const ArrowRight = styled.div`
   }
 `
 
+export const IconLeft = styled(Icon)`
+  margin-right: 10px;
+`
+export const IconRight = styled(Icon)`
+  margin-left: 10px;
+`
+
 const getVariant = () => ({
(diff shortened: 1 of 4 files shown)


packages/asc-ui/src/components/Button/ButtonStyle.ts
```diff
@@ -115,14 +115,29 @@ const getVariant = () => ({
...
         &:hover {
           color: ${themeColor('secondary')};
           ${svgFill('secondary')};
         }
+
+        ${IconLeft} {
+          margin-right: 5px;
+        }
+        ${IconRight} {
+          margin-left: 5px;
```
chrisvanmook: can u use `themeSpacing` util here? As we use a unit of 4px for margins

packages/asc-ui/src/components/Button/ButtonStyle.ts
```diff
@@ -241,19 +256,6 @@ const ButtonStyle = styled.button<Props>`
         opacity: 1;
       }
     `}
-  ${({ variant }) =>
```
royderks: Why move this styling up? Looked like it's only applied to the `:disabled` state of the button

PR #191, opened 2019-10-31
Added active prop to MenuButton to enable activeStyles

robinpiets:


packages/asc-ui/src/components/Button/Button.tsx
```diff
@@ -16,6 +16,7 @@ export type Props = {
   iconSize?: number
   $as?: any
   taskflow?: boolean
+  active?: boolean
```
royderks: Would add this at the level of MenuButton, as I dont expect we want to set this prop on a Button directly

robinpiets: Fixed. Thanks for pointing me in the right direction

PR #201, opened 2019-11-14
Make SearchBar button optional and allow for custom button icon.

spaces-ga:
This PR would allow:
- to hide the search button on the SearchBar component
- to supply a different search button icon to the SearchBar component


royderks:
Would prefer having a `handler` prop just as for the `Toggle` component, the default value could be the `Button` as used in this code. If passed `null` no button would be displayed, if passed nothing the default button and if any other button it would render that. What do you think @spaces-ga ?

spaces-ga:
@royderks Sounds like a valid solution. Will look into that.

royderks:
Do you still want to complete this PR using the hinted solution @spaces-ga?

jonkoops:
Closing this due to inactivity.

PR #202, opened 2019-11-14
Expose context providers.

spaces-ga:
This PR exposes the Context Providers used by various components. This way we could easily create our own custom versions of components.


royderks:
Can you give an example of what you like to accomplish in your project with this PR? @spaces-ga

spaces-ga:
@royderks The `Input` component consumers `InputMethodContext`. This is great for creating your own custom or advanced textfields/searchbars etc, But only if you could actually set anything on the `InputMethodContext`. Currently this is not possible as this Provider is not exposed.

royderks:
If you want to do a canary release to test this locally you can run the command yarn publish:canary from your branch. It's important to do this before pushing to the master branch @spaces-ga

PR #205, opened 2019-12-05
(chore) Remove hardcoded input attributes

janjaap:
This PR removes attributes that were hardcoded on the `Input` component. These attributes prevented setting the `type` attribute of the HTMLElement, for instance.


chrisvanmook:
@janjaap tests are still failing, prettier stuff...

jonkoops:
I've gone ahead and rebased this PR with master to remove all the weird commits. @janjaap can you give this a check and see if everything as it should be?

PR #220, opened 2020-01-15
Add CompactPager component

jonkoops:
This PR adds a new CompactPager component which allows the user to create a pager for a collection of a set number of items and divide them into a specified number of pages.

image

### DIFF
--- a/packages/asc-ui/src/components/CompactPager/CompactPager.test.tsx
+++ b/packages/asc-ui/src/components/CompactPager/CompactPager.test.tsx
@@ -1,0 +1,74 @@
+import * as React from 'react'
+import { render, fireEvent } from '@testing-library/react'
+import { ThemeProvider, ascDefaultTheme } from '@datapunt/asc-core'
+import CompactPager, { determineTotalPages } from './CompactPager'
+import { PREV_BUTTON, NEXT_BUTTON } from './test-ids'
+
+describe('CompactPager', () => {
+  const MockComponent = (props: any) => (
+    <ThemeProvider theme={ascDefaultTheme}>
+      <CompactPager page={1} pageSize={20} collectionSize={60} {...props} />
+    </ThemeProvider>
+  )
+
+  it('should change the page when pressing prev button', () => {
+    const onPageChangeMock = jest.fn()
+    const { getByTestId } = render(
+      <MockComponent page={2} onPageChange={onPageChangeMock} />,
+    )
+
+    fireEvent.click(getByTestId(PREV_BUTTON))
+    expect(onPageChangeMock).toHaveBeenCalledWith(1)
+  })
+
+  it('should not change the page at the bounds when pressing the prev button', () => {
+    const onPageChangeMock = jest.fn()
+    const { getByTestId } = render(
+      <MockComponent page={1} onPageChange={onPageChangeMock} />,
+    )
+
+    fireEvent.click(getByTestId(PREV_BUTTON))
+    expect(onPageChangeMock).not.toHaveBeenCalled()
+  })
+
+  it('should change the page when pressing next button', () => {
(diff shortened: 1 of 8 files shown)


packages/asc-ui/src/components/CompactPager/CompactPager.tsx
```diff
@@ -0,0 +1,93 @@
+import React from 'react'
+import { ChevronRight, ChevronLeft } from '@datapunt/asc-assets'
+import CompactPagerStyle, { PagerText, PagerButton } from './CompactPagerStyle'
+
+function clampNumber(value: number, min: number, max: number) {
+  return Math.min(Math.max(value, min), max)
+}
+
+export function determineTotalPages(collectionSize: number, pageSize: number) {
```
royderks: Can't we use the props as defined below for `collectionSize` and `pagesize`? Maybe move these two functions two a `utils.ts` and define the props there

packages/asc-ui/src/components/CompactPager/CompactPager.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    <CompactPagerStyle {...otherProps}>
+      <PagerButton
+        size={44}
+        variant="blank"
+        icon={<ChevronLeft />}
+        disabled={!hasPrevPage}
+        aria-label="Vorige pagina"
+        onClick={() => movePageCursor(-1)}
+        data-testid="prevbutton"
```
royderks: You can define these in constants, just as you did in the test file

packages/asc-ui/src/components/CompactPager/CompactPager.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { render, fireEvent } from '@testing-library/react'
+import { ThemeProvider, ascDefaultTheme } from '@datapunt/asc-core'
+import CompactPager, { determineTotalPages } from './CompactPager'
+
+const PREV_BUTTON = 'prevbutton'
+const NEXT_BUTTON = 'nextbutton'
+
+describe('CompactPager', () => {
+  it('should change the page when pressing prev button', () => {
+    const onChangePageMock = jest.fn()
```
royderks: You're setting the same mock function and mock component four times, defining it just once is enough

jonkoops: The arguments passed in are different sometimes, but yeah, I'll add it to a function so it can be re-used.

packages/asc-ui/src/components/CompactPager/CompactPager.tsx
jonkoops: Not quite sure how to optimize this to be honest. I think it's fine for these functions to live next to the component like this since they are exclusively used there.

We could pass the entire props to this method, but then we also pass a lot of information that this function doesn't need, also complicating the tests more.

packages/asc-ui/src/components/CompactPager/CompactPager.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    )
+
+    fireEvent.click(getByTestId(PREV_BUTTON))
+    expect(onChangePageMock).toHaveBeenCalledWith(1)
+  })
+
+  it('should not change the page at the bounds when pressing the prev button', () => {
+    const onChangePageMock = jest.fn()
+
+    const { getByTestId } = render(
```
chrisvanmook: Perhaps create a createComponent function with props as arguments to make tests a bit more compact / less repetitive?

jonkoops: Yes, I pushed a change that should resolve this.

PR #294, opened 2020-03-03
Add a forwardRef to the Input component

jasperswart:
In this PR:
 - added ref to Input component
- added test

### DIFF
--- a/packages/asc-ui/src/components/Input/Input.tsx
+++ b/packages/asc-ui/src/components/Input/Input.tsx
@@ -1,8 +1,8 @@
-import React from 'react'
-import InputStyle from './InputStyle'
-import InputContext from './InputMethodsContext'
-import useFocusOnRender from '../../utils/hooks/useFocusOnRender'
+import React, { useImperativeHandle } from 'react'
 import useActionOnEscape from '../../utils/hooks/useActionOnEscape'
+import useFocusOnRender from '../../utils/hooks/useFocusOnRender'
+import InputContext from './InputMethodsContext'
+import InputStyle from './InputStyle'
 
 export interface InputMethods {
   onBlur?: (event: React.FocusEvent<HTMLInputElement>) => void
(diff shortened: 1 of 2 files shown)


packages/asc-ui/src/components/Input/Input.tsx
```diff
@@ -23,62 +23,63 @@ export interface InputProps extends InputMethods {
   error?: string
 }
 
-export interface Shared {}
+const Input = React.forwardRef<
+  HTMLInputElement,
+  InputProps & React.HTMLAttributes<HTMLInputElement>
+>(
+  (
+    { blurOnEscape, focusOnRender, value, error, ...props }: InputProps,
```
chrisvanmook: If i'm not mistaken, you can remove the  :InputProps and the :any, as these should already be typed

jasperswart: dankjewel

jasperswart: gedaan

packages/asc-ui/src/components/Input/Input.tsx
```diff
@@ -28,7 +28,16 @@ const Input = React.forwardRef<
...
   const internalRef = React.useRef<HTMLInputElement>(null)
+  React.useImperativeHandle<HTMLInputElement, any>(externalRef, () => ({
+    blur: () => {
+      if (internalRef.current) {
+        internalRef.current.blur()
+      }
+    },
+  }))
+
   const ref = externalRef || internalRef
```
chrisvanmook: So this line can be removed, and `internalRef` can be renamed to `ref`

jonkoops:
Looks like we we're using `useImperativeHandle` in the wrong manner, I've changed the implementation and squashed the commits. Tests seem to pass properly now as they are indeed recieving a `HTMLInputElement` as expected.

Issue #11, opened 2019-04-04
Add component description


aditudorache:
- Add for each component a (short)description that makes clear what the component does and eventually some usage scenarios if needed

jonkoops:
We're going to add this for every component we stabilize.

PR #301, opened 2020-03-05
Add the TopTaskButton component

jasperswart:
In this PR:
- added top task button

### DIFF
--- a/packages/asc-ui/src/components/TopTaskButton/TopTaskButton.stories.tsx
+++ b/packages/asc-ui/src/components/TopTaskButton/TopTaskButton.stories.tsx
@@ -1,0 +1,55 @@
+import { Location, Passport } from '@datapunt/asc-assets'
+import React from 'react'
+import Readme from './README.md'
+import TopTaskButton from './TopTaskButton'
+
+export default {
+  title: 'Experimental/Atoms/TopTaskButton',
+  parameters: {
+    notes: Readme,
+  },
+  decorators: [
+    (storyFn: () => React.ReactNode) => (
+      <div
+        style={{
+          display: 'flex',
+          maxWidth: 400,
+          flexDirection: 'column',
+          alignItems: 'flex-start',
+          padding: '40px 10px',
+        }}
+      >
+        {storyFn()}
+      </div>
+    ),
+  ],
+}
+
+export const WithIcon = () => (
+  <TopTaskButton
+    href="https://amsterdam.nl"
+    icon={Location}
+    text="Stadsloket: locaties en openingstijden"
+    style={{ maxWidth: 220 }}
+  />
+)
+
+export const WithoutIcon = () => (
+  <TopTaskButton
+    href="https://amsterdam.nl"
+    text="Stadsloket: locaties en openingstijden"
+    style={{ maxWidth: 220 }}
+  />
+)
+
+export const OnDarkBackground = () => (
+  <div style={{ backgroundColor: '#004699', padding: '40px 80px' }}>
+    <TopTaskButton
+      href="https://amsterdam.nl"
+      target="_blank"
+      icon={Passport}
+      text="Stadsloket: locaties en openingstijden"
+      onDarkBackground
+    />
+  </div>
+)
--- a/packages/asc-ui/src/components/TopTaskButton/TopTaskButton.tsx
+++ b/packages/asc-ui/src/components/TopTaskButton/TopTaskButton.tsx
@@ -1,0 +1,30 @@
+import React, { AnchorHTMLAttributes } from 'react'
+import Icon from '../Icon'
+import TopTaskButtonStyle, { ParagraphStyle } from './TopTaskButtonStyle'
+
+export interface TopTaskButtonProps {
+  icon?: React.ElementType
+  text: string
+  onDarkBackground?: boolean
+}
+
+const TopTaskButton: React.FC<
+  TopTaskButtonProps & AnchorHTMLAttributes<HTMLAnchorElement>
+> = ({ icon, text, ...otherProps }) => {
+  const CustomIcon = icon
+
+  return (
+    <TopTaskButtonStyle {...otherProps} data-testid="root">
+      {CustomIcon && (
+        <Icon inline size={32} padding={0} data-testid="icon">
+          <CustomIcon />
+        </Icon>
+      )}
+      <ParagraphStyle styleAs="h4" strong>
+        {text}
+      </ParagraphStyle>
+    </TopTaskButtonStyle>
+  )
+}
+
+export default TopTaskButton
--- a/packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
+++ b/packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
@@ -1,0 +1,35 @@
+import styled from 'styled-components'
+import { svgFill, themeColor, themeSpacing } from '../../utils'
+import IconStyle from '../Icon/IconStyle'
+import Paragraph from '../Paragraph'
+
+export default styled.a<{
+  onDarkBackground?: boolean
+}>`
+  display: block;
+  padding: ${themeSpacing(2, 3)};
+  text-decoration: none;
+  color: ${themeColor('tint', 'level7')};
+  background-color: ${themeColor('tint', 'level2')};
+  border-style: solid;
+  border-width: 0 2px 2px 0;
+  border-color: ${({ onDarkBackground = false }) =>
+    onDarkBackground
+      ? themeColor('tint', 'level7')
+      : themeColor('tint', 'level3')};
+
+  :hover {
+    border-color: ${themeColor('secondary', 'main')};
+    color: ${themeColor('secondary', 'main')};
+    text-decoration: underline;
+  }
+
+  ${IconStyle} {
+    margin-bottom: ${themeSpacing(2)};
+    ${svgFill('tint', 'level7')};
+  }
+`
+
+export const ParagraphStyle = styled(Paragraph)`
+  margin-bottom: 0;
+`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,14 +16,13 @@ Prefix the change with one of these keywords:
 
 ## Unreleased
 - Changed: Internal SVG icon imports are replaced with React components, eliminating the need for including an SVG loader (closes #459)
-
+- Added: Added TopTaskButton component
 - Added: `theme` prop to `ThemeProvider` to pass a custom theme
 - Changed: **BREAKING** `createTheme` method from `Theme`.
 - Changed: **BREAKING** removed `themeName` from `ThemeProvider`
 - Changed: **BREAKING** removed `asc-core` package. `Theme` and `ascDefaultTheme` can now be found in `asc-ui`
 
 ## [0.19.2-beta.1]
-
 - Added Hidden component
 
 ## [0.19.1]
--- a/packages/asc-ui/src/components/TopTaskButton/README.md
+++ b/packages/asc-ui/src/components/TopTaskButton/README.md
@@ -1,0 +1,11 @@
+# TopTask button
+
+Used as a navigational element to direct a user to a specific subject.
+[Design System](https://designsystem.amsterdam.nl/7awj1hc9f/p/46e2d8-toptask-button)
+
+**Component status: experimental**
+
+## Todo's before status "stable"
+
+- Discuss the open points on the Trello board with the design system team.
+
--- a/packages/asc-ui/src/components/TopTaskButton/TopDownButton.test.tsx
+++ b/packages/asc-ui/src/components/TopTaskButton/TopDownButton.test.tsx
@@ -1,0 +1,28 @@
+import { Location } from '@datapunt/asc-assets'
+import { render } from '@testing-library/react'
+import React from 'react'
+import { ThemeProvider } from '../../theme'
+import TopTaskButton from './TopTaskButton'
+
+describe('TopTaskButton', () => {
+  it('should render', () => {
+    const { queryByText, queryByTestId } = render(
+      <ThemeProvider>
+        <TopTaskButton
+          href="https://amsterdam.nl"
+          icon={Location}
+          text="Stadsloket: locaties en openingstijden"
+        />
+      </ThemeProvider>,
+    )
+
+    expect(
+      queryByText('Stadsloket: locaties en openingstijden'),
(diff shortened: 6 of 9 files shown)


jonkoops:
@jasperswart Thanks for the PR, I'll take a look as soon as I have some time.

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+      >
+        {storyFn()}
+      </div>
+    ),
+  ],
+}
+
+export const OnWhiteBackgroundWithAnchorTag = () => (
+  <>
+    <a href="https://amsterdam.nl" target="_blank" rel="noopener noreferrer">
```
chrisvanmook: Shouldn't TopTaskButton itself be a Button or a Link as a default? Now we need to add css on the div inside for the hover style, that should be on a button or link imho. Then TopTaskButton should be a Button as a default, where you could pass all button (html)props. @jonkoops what do you think?

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+const defaultProps = {
+  onDarkBackground: false,
+  width: '220px',
+}
+
+export default styled.div<Props>`
+  width: ${({ width = defaultProps.width }) => width};
+  padding: ${themeSpacing(2)} ${themeSpacing(4)};
```
chrisvanmook: Can be written like `padding: ${themeSpacing(2, 4)};`

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+  ...otherProps
+}) => {
+  const CustomIcon = icon
+
+  return (
+    <TopTaskButtonStyle {...otherProps}>
+      <Icon inline size={32} color="black">
+        <CustomIcon />
+      </Icon>
+      <div>{title}</div>
```
chrisvanmook: Can we use `<Paragraph strong>` here?

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+      <section>{content}</section>
```
chrisvanmook: Why the section? Can we use `<Paragraph>` here?

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
```diff
@@ -0,0 +1,48 @@
...
+export interface Props {
+  icon: React.ElementType
+  title?: string
+  content?: string
+  onDarkBackground?: boolean
+  width: string
+  className: string
+}
+
+const defaultProps = {
```
chrisvanmook: Perhaps we can move the defaultProps to the .tsx file? Then we don't need to use it inside the styled-component.

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import TopTaskButtonStyle, { Props } from './TopTaskButtonStyle'
+import Icon from '../Icon'
+
+const TopTaskButton: React.FC<Props> = ({
```
chrisvanmook: Then consider to use a Button or Link as a default, we can type this like:
`React.FC<Props & React.AnchorHTMLAttributes<HTMLAnchorElement> &  React.ButtonHTMLAttributes<HTMLButtonElement>>`

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import IconStyle from '../Icon/IconStyle'
+
+export interface Props {
+  icon: React.ElementType
+  title?: string
+  content?: string
+  onDarkBackground?: boolean
+  width: string
+  className: string
```
chrisvanmook: This is already typed when you type the React component in the tsx file (see previous comment)

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+const defaultProps = {
+  onDarkBackground: false,
+  width: '220px',
+}
+
+export default styled.div<Props>`
+  width: ${({ width = defaultProps.width }) => width};
```
chrisvanmook: I think this shouldn't be here. the user of the component should decide the width of the toptask (either by wrapping it in a grid, or extend it like `styled(TopTaskButton)'width: 300px;'`)

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
```diff
@@ -0,0 +1,48 @@
...
+  padding: ${themeSpacing(2)} ${themeSpacing(4)};
+  cursor: pointer;
+  font-weight: bold;
+  font-size: 18px;
```
chrisvanmook: We can remove the font-size / weight properties when we use the Paragraph component

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  :hover {
+    box-shadow: 2px 2px ${themeColor('secondary', 'main')};
+    color: ${themeColor('secondary', 'main')};
+    text-decoration: underline;
+  }
+
+  ${IconStyle} {
+    margin: 0;
+    padding: 0 0 10px 1px;
```
chrisvanmook: Why the 1px? And can we use themeSpacing here?

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.stories.tsx
jasperswart: yes I agree, but what to do with redux links?

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
jasperswart: done

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
jasperswart: it issued for the width and onDarkBackground

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.tsx
jasperswart: we cannot make it an anchor because otherwise you cannot use Router Links, but we can use a Button instead

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
jasperswart: you are right, I will remove that

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
jasperswart: no it needs to be 1 pixel

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
jasperswart: done

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.tsx
jasperswart: it is a button now

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
jasperswart: done

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
jasperswart: done

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.tsx
jasperswart: done

packages/asc-ui/src/components/Grid/Grid.stories.tsx
```diff
@@ -5,7 +5,7 @@ import Column from './Column'
 import Row from './Row'
 
 export default {
-  title: 'Atoms/Grid',
+  title: 'Deprecated/Grid',
```
chrisvanmook: This is confusing, but components `Column` and `Row` are part of the story `Grid`, while the grid component is deprecated. So this can be discarded

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.stories.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
```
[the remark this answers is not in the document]
chrisvanmook: This has been a problem for a while now. Storybook supports .mdx files, but is is (or was at least) a hell to set that up with typescript. https://github.com/Amsterdam/amsterdam-styled-components/issues/11

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.stories.tsx
chrisvanmook: Perhaps use `forwardedAs` with other props?

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
chrisvanmook: Yes but why? :P

jonkoops:
I've given the PR a quick rebase and squash so we have a clean commit history. I'll add some commits tomorrow to add some feedback.

packages/asc-ui/src/components/Grid/Grid.stories.tsx
jonkoops: Yeah this came from the rebase

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export const WithoutIcon = () => (
+  <TopTaskButton
+    href="https://amsterdam.nl"
+    text="Stadsloket: locaties en openingstijden"
+    style={{ maxWidth: 220 }}
+  />
+)
+
+export const OnDarkBackground = () => (
```
royderks, line 45: Can we add an example with an onclick event?

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
```diff
@@ -0,0 +1,35 @@
+import styled from '@datapunt/asc-core'
+import { svgFill, themeColor, themeSpacing } from '../../utils'
+import IconStyle from '../Icon/IconStyle'
+import Paragraph from '../Paragraph'
+
+export default styled.a<{
```
royderks, line 6: Why an anchor and not a button?

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+  onDarkBackground?: boolean
+}
+
+const TopTaskButton: React.FC<
+  TopTaskButtonProps & AnchorHTMLAttributes<HTMLAnchorElement>
+> = ({ icon, text, ...otherProps }) => {
+  const CustomIcon = icon
+
+  return (
+    <TopTaskButtonStyle {...otherProps} data-testid="root">
```
royderks, line 17: Should we pass a `title` prop?

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.stories.tsx
jonkoops: The intention is that this should always be an anchor element, since this is a navigational element.

As far as the href is concerned it would just be an empty link right? Which is something you can also do in plain HTML, even if it's not the best idea.

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
jonkoops: Since the TopTaskButton is always used as a navigational element we decided to use an anchor element.

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.tsx
jonkoops: We could, but for what purpose?

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
royderks: Ok, so why not rename it to TopTaskLink? Button is very confusing if it's not a button but a link..

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.tsx
royderks: I'm not sure the `text` prop will always be used as the anchor description, as the CustomIcon could be an element with an `alt` tag right?

jonkoops: It should take the contents of the anchor element as it's description, so in this case the text in the body (.e.g the `text` prop). As far as the icon is concerned it is a decorative icon so it should not be read by screen readers.

packages/asc-ui/src/components/TopTaskButton/TopTaskButtonStyle.ts
jonkoops: I'll propose it at the design system meeting, the terminology is taken from there.

PR #362, opened 2020-03-25
Added indeterminate state to Checkbox

chrisvanmook:

### DIFF
--- a/packages/asc-ui/src/components/Checkbox/Checkbox.tsx
+++ b/packages/asc-ui/src/components/Checkbox/Checkbox.tsx
@@ -1,5 +1,11 @@
-import React, { useState, useContext, useMemo } from 'react'
-import { Checkmark } from '@datapunt/asc-assets'
+import React, {
+  useState,
+  useContext,
+  useMemo,
+  useImperativeHandle,
+  useEffect,
+} from 'react'
+import { Checkmark, Indeterminate } from '@datapunt/asc-assets'
 import CheckboxStyle, {
   CheckboxIconStyle,
   CheckboxWrapperStyle,
@@ -19,13 +25,23 @@ const Checkbox = React.forwardRef<
       variant,
       disabled,
       error,
+      indeterminate,
       ...otherProps
     },
-    ref,
+    externalRef,
   ) => {
     const [checked, setChecked] = useState(!!checkedProp)
     const [focus, setFocus] = useState(false)
     const { setActive } = useContext(LabelContext)
+    const ref = React.useRef<HTMLInputElement>(null)
+
+    useImperativeHandle(externalRef, () => ref.current as HTMLInputElement)
+
+    useEffect(() => {
+      if (ref.current) {
+        ref.current.indeterminate = indeterminate ?? false
+      }
+    }, [ref, indeterminate])
 
     // Make the label aware of changes in the checked state
     useMemo(() => {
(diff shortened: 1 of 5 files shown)


packages/asc-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -40,10 +41,12 @@ const Checkbox = React.forwardRef<
     return (
       <CheckboxWrapperStyle
         {...{ className, disabled, focus, checked, error }}
+        // eslint-disable-next-line no-nested-ternary
+        aria-checked={checked ? (indeterminate ? 'mixed' : true) : false}
```
jonkoops: Do we need to add these ARIA attributes? `CheckboxStyle` is already an `<input type="checkbox">`, it should provide all the information that is needed here already. It also has support for setting the `indeterminate` state, see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Indeterminate_state_checkboxes

packages/asc-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -40,13 +41,14 @@ const Checkbox = React.forwardRef<
     return (
       <CheckboxWrapperStyle
         {...{ className, disabled, focus, checked, error }}
+        // eslint-disable-next-line no-nested-ternary
         aria-disabled={disabled}
```
jonkoops: Do we need to add this `aria-disabled` attribute here? Since the checkbox itself already gets the disabled state as well.

packages/asc-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -40,13 +41,14 @@ const Checkbox = React.forwardRef<
...
       >
         <CheckboxIconStyle {...{ disabled, variant, checked, error }} size={15}>
-          {checked && <Checkmark />}
+          {checked && (indeterminate ? <Indeterminate /> : <Checkmark />)}
         </CheckboxIconStyle>
         <CheckboxStyle
-          {...{ ...otherProps, disabled, checked, ref }}
+          {...{ ...otherProps, disabled, checked, indeterminate, ref }}
```
jonkoops: `indeterminate` is a property on `HTMLInputElement`, not an attribute. This means that you cannot simply set it as a prop, you will have to use the `ref` of the element and set it there manually. For more information check out this blog [post](https://davidwalsh.name/react-indeterminate
).

chrisvanmook: This one is not meant to be set as an attribute, but needed this for styling at first, but it's not needed anymore so I removed it :)

packages/asc-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -19,13 +25,23 @@ const Checkbox = React.forwardRef<
...
     const [focus, setFocus] = useState(false)
     const { setActive } = useContext(LabelContext)
+    const ref = React.useRef<HTMLInputElement>(null)
+
+    useImperativeHandle(externalRef, () => ref.current as HTMLInputElement)
+
+    useEffect(() => {
+      if (ref.current && typeof indeterminate === 'boolean') {
+        ref.current.indeterminate = indeterminate
+      }
```
jonkoops: If the `indeterminate` prop was `true` and then becomes `undefined` it will stay `true` since the type is checked. Best to default to `false` here if `indeterminate` is not set.

```suggestion
      if (ref.current) {
        ref.current.indeterminate = indeterminate ?? false
      }
```

chrisvanmook: Nice one, I should start to use this operator more :)

PR #490, opened 2020-04-23
Implement ErrorMessage component

jasperswart:
Added ErrorMessage component

Used this spec: https://designsystem.amsterdam.nl/7awj1hc9f/p/35e7b1-error-message

### DIFF
--- a/packages/asc-ui/src/components/ErrorMessage/ErrorMessage.stories.tsx
+++ b/packages/asc-ui/src/components/ErrorMessage/ErrorMessage.stories.tsx
@@ -1,0 +1,80 @@
+import React from 'react'
+import { Checkbox } from '../Checkbox'
+import Input from '../Input'
+import Label from '../Label'
+import Radio from '../Radio'
+import RadioGroup from '../Radio/RadioGroup'
+import Select from '../Select'
+import ErrorMessage from './ErrorMessage'
+import Readme from './README.md'
+
+export default {
+  title: 'Experimental/Atoms/ErrorMessage',
+  parameters: {
+    notes: Readme,
+  },
+  decorators: [
+    (storyFn: () => React.ReactNode) => (
+      <div
+        style={{
+          display: 'flex',
+          maxWidth: 400,
+          flexDirection: 'column',
+          alignItems: 'flex-start',
+          padding: '40px 10px',
+        }}
+      >
+        {storyFn()}
+      </div>
+    ),
+  ],
+}
+
+export const Default = () => (
+  <ErrorMessage message="This is an incorrect value" />
+)
+
+export const WithInput = () => (
+  <>
+    <Input error />
+    <ErrorMessage message="This a required field" />
+  </>
+)
+
+export const WithRadio = () => (
+  <>
+    <RadioGroup name="group-1" error>
+      <Label htmlFor="radio-1" label="Windows">
+        <Radio id="radio-1" />
+      </Label>
+      <Label htmlFor="radio-2" label="MacOS">
+        <Radio id="radio-2" />
+      </Label>
+      <Label htmlFor="radio-3" label="Linux">
+        <Radio id="radio-3" />
+      </Label>
+    </RadioGroup>
+    <ErrorMessage message="This a required field" />
+  </>
+)
+
+export const WithCheckbox = () => (
+  <>
+    <Label htmlFor="error" label="Accept terms">
+      <Checkbox id="error" error />
+    </Label>
+    <ErrorMessage message="This a required field" />
+  </>
+)
+
+export const WithSelect = () => (
+  <>
+    <Select id="select-1" error>
+      <option value="1">Option 1</option>
+      <option value="2">Option 2</option>
+      <option value="3">Option 3</option>
+      <option value="4">Option 4</option>
+    </Select>
+    <ErrorMessage message="This a required field" />
+  </>
+)
--- a/packages/asc-ui/src/components/ErrorMessage/ErrorMessage.test.tsx
+++ b/packages/asc-ui/src/components/ErrorMessage/ErrorMessage.test.tsx
@@ -1,0 +1,19 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+import { ThemeProvider } from '../../theme'
+import ErrorMessage from './ErrorMessage'
+
+describe('ErrorMessage', () => {
+  it('should render', () => {
+    const { container } = render(
+      <ThemeProvider>
+        <ErrorMessage message="This is an incorrect value" />
+      </ThemeProvider>,
+    )
+
+    expect(container.firstChild).toBeInTheDocument()
+    expect(container.firstChild?.textContent).toEqual(
+      'This is an incorrect value',
+    )
+  })
+})
--- a/packages/asc-ui/src/components/ErrorMessage/ErrorMessage.tsx
+++ b/packages/asc-ui/src/components/ErrorMessage/ErrorMessage.tsx
@@ -1,0 +1,18 @@
+import React, { HTMLAttributes } from 'react'
+import ErrorMessageStyle from './ErrorMessageStyle'
+
+export interface ErrorMessageProps {
+  message: string
+}
+
+const ErrorMessage: React.FC<
+  ErrorMessageProps & HTMLAttributes<HTMLDivElement>
+> = ({ message, ...otherProps }) => {
+  return (
+    <ErrorMessageStyle role="alert" {...otherProps}>
+      {message}
+    </ErrorMessageStyle>
+  )
+}
+
+export default ErrorMessage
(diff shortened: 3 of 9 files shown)


packages/asc-ui/src/components/TopTaskButton/TopTaskButton.stories.tsx
```diff
@@ -1,13 +1,9 @@
 import { Location, Passport } from '@datapunt/asc-assets'
 import React from 'react'
-import Readme from './README.md'
```
jonkoops, line 3: I think these are some changes that were reverted by accident. Try pulling the latest version from master using rebase.

yarn.lock
```diff
@@ -4017,11 +4017,6 @@ acorn@^6.0.1, acorn@^6.4.1:
   resolved "https://registry.yarnpkg.com/acorn/-/acorn-6.4.1.tgz#531e58ba3f51b9dacb9a6646ca4debf5b14ca474"
```
jonkoops, line 4017: Strange the lockfile has changed, since no deps have changed.

packages/asc-ui/src/components/TopTaskButton/TopTaskButton.stories.tsx
jasperswart: there was an incorrect change in the topTaskButton: I have reverted the change

packages/asc-ui/src/components/ErrorMessage/ErrorMessage.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+      >
+        {storyFn()}
+      </div>
+    ),
+  ],
+}
+
+export const Default = () => <ErrorMessage message="Dit is dus fout" />
+
+export const NoMessage = () => <ErrorMessage />
```
robinpiets, line 25: Hi @jasperswart, can you add stories to show how it looks when used with for example `Checkbox`, `Radio`, and/or `TextField` components?

packages/asc-ui/src/components/ErrorMessage/ErrorMessage.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+          padding: '40px 10px',
+        }}
+      >
+        {storyFn()}
+      </div>
+    ),
+  ],
+}
+
+export const Default = () => <ErrorMessage message="Dit is dus fout" />
```
robinpiets, line 23: Nitpicking: this package is using only English text throughout the app

packages/asc-ui/src/components/ErrorMessage/ErrorMessage.test.tsx
```diff
@@ -0,0 +1,28 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+import { ThemeProvider } from '../../theme'
+import ErrorMessage from './ErrorMessage'
+
+describe('ErrorMessage', () => {
+  it('should render correctly', () => {
+    const { queryByText, queryByTestId } = render(
+      <ThemeProvider>
+        <ErrorMessage message="Dit veld is verplicht" />
```
robinpiets, line 10: Nitpicking: this package is using only English text throughout the app

jasperswart: gedaan

yarn.lock
jasperswart: I do not understand why

packages/asc-ui/src/components/ErrorMessage/ErrorMessage.stories.tsx
jasperswart: gedaan

packages/asc-ui/src/components/ErrorMessage/ErrorMessage.stories.tsx
jasperswart: I have added the examples for input, select, radio and checkbox to story

jasperswart: can you check?

robinpiets: I like it very much

yarn.lock
robinpiets: Off topic: Any particular reason not to drop `yarn` and choose `npm`?

packages/asc-ui/src/components/ErrorMessage/ErrorMessage.test.tsx
```diff
@@ -0,0 +1,28 @@
+import { render } from '@testing-library/react'
+import React from 'react'
+import { ThemeProvider } from '../../theme'
+import ErrorMessage from './ErrorMessage'
+
+describe('ErrorMessage', () => {
+  it('should render correctly', () => {
+    const { queryByText, queryByTestId } = render(
+      <ThemeProvider>
+        <ErrorMessage message="This a an incorrect value" />
```
robinpiets, line 10: typo, but it's a consistent typo, so who cares?

jasperswart: sorry, weg nu

yarn.lock
chrisvanmook: @robinpiets yarn supports workspaces. https://doppelmutzi.github.io/monorepo-lerna-yarn-workspaces/

chrisvanmook: @jasperswart Perhaps it's good thing to discard these changes

packages/asc-ui/src/components/ErrorMessage/ErrorMessageStyle.ts
```diff
@@ -0,0 +1,11 @@
+import styled from 'styled-components'
+import { themeColor } from '../../utils'
+
+const ErrorMessageStyle = styled.div`
+  color: ${themeColor('error', 'main')};
+  font-weight: 700;
+  line-height: 22px;
+  margin-top: 15px;
```
chrisvanmook, line 8: rather use `themeSpacing(4)` here

yarn.lock
chrisvanmook: or we can merge https://github.com/Amsterdam/amsterdam-styled-components/pull/505 and rebase these changes onto master, so netlify can build :)

jonkoops:
@robinpiets Yeah, I agree there are significant problems with these existing components. For example the `TextField` component tries to take care of a lot of things and ends up being a bit bloated.

Note that in multiple places the design system is also inconsistent about where labels are placed, or what margins are used for spacing. We need to discuss this with the Design System team as well. 

There are also various A11Y issues we need to sort out around the error state. For example, how does the browser know that a field in in fact invalid? Or how does the screen reader know which error message belongs to which field? These are all questions we will have to answer.

Personally for the next step I would advocate we remove the label/error message functionality from the `Select` and `TextField` component. This seems like a component taking on too much functionality when it could just be composed by the end-user.

I'd like to end up with something we can compose together like:
```tsx
<Label for="select-1">Select an option</Label>
<Select id="select-1">
  <option value="1">Option 1</option>
  <option value="2">Option 2</option>
  <option value="3">Option 3</option>
  <option value="4">Option 4</option>
</Select>
{errors.outOfStock && (
  <ErrorMessage for="select-1">This item is out of stock</ErrorMessage>
)}
```

Perhaps we could even introduce a `FormGroup` like element to contain the elements, and reduce some of the boilerplate by providing the `id` to wire things up automatically behind the scenes (accessible by default).

As for this individual component, I think we can merge this PR and introduce it (experimental of course) and fit it onto place of the other components later. I can create some issues to track this.

@chrisvanmook @royderks what do you think?

jonkoops:
For now I will approve this PR, I think we can move on and work the implementation our further.

chrisvanmook:
@jonkoops @robinpiets Indeed, components like TextField are indeed bloated and take care of too many things... A `FormGroup` component would be a good solution to deal with layout like material-ui does https://material-ui.com/api/form-group/#formgroup-api Also it's nice to have the elements wrapped inside something.

I think we can learn a lot from how material and ant design solve these kind of problems

robinpiets:
Good points @jonkoops and @chrisvanmook. Can you make a new Issue for this so we don't forget about it. I can start with this one next week:

_remove the label/error message functionality from the Select and TextField component._

packages/asc-ui/src/components/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export interface ErrorMessageProps {
+  message: string
+}
+
+const ErrorMessage: React.FC<
+  ErrorMessageProps & HTMLAttributes<HTMLDivElement>
+> = ({ message, ...otherProps }) => {
+  return (
+    <ErrorMessageStyle role="alert" data-testid="root" {...otherProps}>
```
aditudorache, line 12: Why is the data-testid called `root`? How about `errorMessage`? So it will have more meaning when used in the implementations.

jonkoops: Good catch, we're not actually using this test id anymore so I removed it.

PR #495, opened 2020-04-24
Add examples of using refs on form elements

### LINKED ISSUE #218: None of the basic form element can be references by `ref`
**Describe the bug**

For any form elements (radiobutton, checkbox, ...) it can be very handy, or at times even a requirement, to have a reference to an element. This is currently not possible with the components that are exposed in this lib.

**To Reproduce**
1. Create a component:
    ```jsx
    import React, { useRef } from 'react';
    import { Checkbox } from '@datapunt/asc-ui';
    
    export const Form = () => {
      const cbRef = useRef(null);
    
      return (
        <form>
          <Checkbox ref={cbRef} />
        </form>
      );
    };
    ```
2. Observe the following error in the dev console:
    > Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

**Expected behaviour**
No errors/warnings in the console

jonkoops:
Closes #218

### DIFF
--- a/packages/asc-ui/src/components/Radio/Radio.stories.tsx
+++ b/packages/asc-ui/src/components/Radio/Radio.stories.tsx
@@ -1,11 +1,14 @@
-import React from 'react'
+import React, { createRef, useEffect } from 'react'
+import Label from '../Label'
 import Radio from './Radio'
 import RadioGroup from './RadioGroup'
-import Label from '../Label'
+import Readme from './README.md'
 
 export default {
   title: 'Experimental/Atoms/Radio',
-
+  parameters: {
+    notes: Readme,
+  },
   decorators: [
     (storyFn: () => React.ReactNode) => (
       <div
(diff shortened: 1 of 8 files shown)


packages/asc-ui/src/components/Radio/Radio.stories.tsx
```diff
@@ -172,3 +175,34 @@ export const RadioLabelPositions = () => (
...
+      console.log('Value changed to option 2!')
+    }
+
+    element.addEventListener('change', onChange)
+
+    return () => element.removeEventListener('change', onChange)
+  }, [ref])
+
+  return (
+    <RadioGroup name="with-ref">
```
royderks, line 196: The usage of id `with-ref` on the group and individual inputs would suggest all elements have a ref, while only the second one has a ref. Can you make this more clear to the user?

jonkoops: Good catch, changed.

PR #509, opened 2020-04-28
Refactor and document useFocus

### LINKED ISSUE #504: Refactor autofocus to simple utility based on refs
Currently some of our components accept a `focusOnRender` prop. Instead of having an implementation of this on each component we want to have a simple utility function that will focus an element based on the ref of the element.

jonkoops:
Closes #504


packages/asc-ui/src/components/TextField/TextField.tsx
```diff
@@ -24,15 +26,19 @@ const TextField = ({
...
   errorMessage,
   labelStyle,
   errorStyle,
   id: idProp,
+  autoFocus,
   ...otherProps
 }: TextFieldProps) => {
   const uid = useUID()
   const id = idProp || uid
+  const inputRef = useRef<HTMLInputElement>(null)
```
robinpiets, line 38: Why do you need `(null)` here?

packages/asc-ui/src/components/SearchBar/SearchBar.tsx
```diff
@@ -20,6 +20,7 @@ export interface SearchBarProps extends TextFieldProps, SearchBarStyleProps {
    * @param value
    */
   onSubmit?: (e: React.FormEvent) => void
+  autoFocus?: boolean
```
robinpiets, line 23: Can you add this functionality to a story of `SearchBar.stories.tsx` as well?

packages/asc-ui/src/components/TextField/TextField.tsx
```diff
@@ -15,6 +16,7 @@ export interface TextFieldProps extends InputProps {
   labelStyle?: object
   errorStyle?: object
   onClear?: () => void
+  autoFocus?: boolean
```
robinpiets, line 19: Can you add this functionality to a story of `TextField.stories.tsx` as well?

packages/asc-ui/src/utils/hooks/useFocus.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect, useRef, useState } from 'react'
+import Button from '../../components/Button'
+import useFocus from './useFocus'
```
robinpiets, line 3: Can you add a README for this? Maybe explain the difference between the `useFocus()` function and `autofocus` prop

packages/asc-ui/src/components/TextField/TextField.tsx
jonkoops: Otherwise we end up with a mutable ref. In this case the ref is defined, just possibly containing a null value. In this case I pulled it from the [React docs](https://reactjs.org/docs/hooks-reference.html#useref).

packages/asc-ui/src/components/SearchBar/SearchBar.tsx
```diff
@@ -30,13 +31,13 @@ const SearchBar: React.FC<SearchBarProps> = ({
   onChange,
   onFocus,
   onWatchValue,
-  focusOnRender,
```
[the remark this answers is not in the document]
jonkoops: I've asked around, nobody seems to be using this at the moment.

packages/asc-ui/src/utils/hooks/useFocus.stories.tsx
jonkoops: There is a story that describes how to use this, and the function itself is documented by JSDoc. That should be enough IMHO

packages/asc-ui/src/components/SearchBar/SearchBar.tsx
jonkoops: Done

packages/asc-ui/src/components/TextField/TextField.tsx
jonkoops: Done

PR #531, opened 2020-05-04
Added new variant to Alert component: warning

robinpiets:
After discussing with Daphnee from DST, I've added a new variant for the Alert component:
image


packages/asc-ui/src/components/Alert/Alert.stories.tsx
```diff
@@ -26,10 +26,14 @@ export const Variants = () => (
     <Alert>Normal</Alert>
     <Alert level="attention">Attention</Alert>
     <Alert level="error">Error</Alert>
+    <Alert level="warning">Error</Alert>
```
jonkoops, line 29: Shouldn't this text be 'Warning'?

packages/asc-ui/src/components/Alert/Alert.stories.tsx
```diff
@@ -26,10 +26,14 @@ export const Variants = () => (
...
     <Alert compact>Normal compact</Alert>
     <Alert compact level="attention">
       Attention compact
     </Alert>
+    <Alert compact level="warning">
+      Error compact
```
jonkoops, line 35: Shouldn't this text be 'Warning compact'?

packages/asc-ui/src/components/Alert/Alert.stories.tsx
```diff
@@ -9,7 +9,7 @@ import Link from '../Link'
```
chrisvanmook, line 9: Perhaps it's also a good idea to create a .mdx file from this, just have a look how I did that for the Checkbox component. Also, because the variants will be included in the Design System, I've created a story per variant for the checkbox, I think it's a good idea to also do that here!

packages/asc-ui/src/components/Alert/Alert.stories.tsx
```diff
@@ -48,7 +52,7 @@ export const CompactState = () => (
       dismissible
       level="attention"
     />
-    <Alert heading="Only heading with button" compact level="error" />
```
jonkoops, line 51: The error alert has been removed here, is that intentional?

packages/asc-ui/src/components/Alert/Alert.stories.tsx
```diff
@@ -82,7 +92,7 @@ export const NormalState = () => (
 
 export const WithCustomChildren = () => (
   <>
-    <Alert level="normal" dismissible>
```
jonkoops, line 85: The error alert has been removed here, is that intentional?

packages/asc-ui/src/components/Alert/AlertStyle.ts
```diff
@@ -91,5 +92,9 @@ export default styled.div<Props>`
           color: ${themeColor('tint', 'level1')};
         }
       `}
+      ${level === 'warning' &&
+      css`
+        border: 2px solid ${themeColor('secondary')};
```
jonkoops, line 97: Does this border change the size of the alert itself? Is that something we should be concerned about?

packages/asc-ui/src/components/Alert/Alert.stories.tsx
robinpiets: Good catch! 🎣

packages/asc-ui/src/components/Alert/Alert.stories.tsx
robinpiets: I just changed `error` to `warning`

packages/asc-ui/src/components/Alert/Alert.stories.tsx
robinpiets: I just changed `<Alert level="normal" dismissible>` to `<Alert level="warning" dismissible>`

packages/asc-ui/src/components/Alert/Alert.stories.tsx
robinpiets: Fixed

packages/asc-ui/src/components/Alert/AlertStyle.ts
robinpiets: It does actually. I'll fix it

packages/asc-ui/src/components/Alert/Alert.stories.mdx
```diff
@@ -0,0 +1,144 @@
...
+  }
+`
+
+<Meta
+  title="Experimental/Atoms/Alert"
+  component={Alert}
+  decorators={[(storyFn) => <Decorator>{storyFn()}</Decorator>]}
+/>
+
+<Description>{Readme}</Description>
```
jonkoops, line 25: I think we can inline the contents of the README here.

packages/asc-ui/src/components/Alert/AlertStyle.ts
robinpiets: Fixed with `box-shadow`

packages/asc-ui/src/components/Alert/Alert.stories.mdx
robinpiets: Done

robinpiets:
Rebased to one commit

PR #538, opened 2020-05-06
Add documentation for icons

jasperswart:
- added documentation for Icon component
- added documentation for Icon list

- fixed issue with icon color incorrectly set when CIRCLE element is in svg
- fixed issue with incorrect handling of prop: iconUrl


packages/asc-ui/src/utils/themeUtils.ts
```diff
@@ -225,6 +225,7 @@ export const svgFill = withTheme<[Theme.ColorType?, string?, string?]>(
       if (typeof value === 'string') {
         return css`
           & svg {
+            circle,
```
jonkoops, line 228: Nice catch!

packages/asc-ui/src/components/IconList/IconList.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+        }}
+      >
+        {storyFn()}
+      </div>
+    ),
+  ],
+}
+
+const allIcons = Object.values(Icons)
+const allLabels = Object.keys(Icons)
```
jonkoops, line 34: Instead of using `Object.values` or `Object.keys` you could use `Object.entries` which will get you a list of both that you can iterate on.

packages/asc-ui/src/utils/themeUtils.ts
jasperswart: yes

PR #539, opened 2020-05-06
Added supplement colors to default theme

alimpens:
Added supplement colors from [design system](https://designsystem.amsterdam.nl/7awj1hc9f/p/63dbfc-colours/t/e4d8) to default theme.


packages/asc-ui/src/theme/default/colors.ts
```diff
@@ -23,6 +23,15 @@ const colors: Theme.ColorTypeInterface = {
     invalid: '#EC0000',
     focus: '#FEC813',
   },
+  supplement: {
+    purple: '#A00078',
```
chrisvanmook, line 27: I'm honestly not a big fan of naming colors to their actual color name. If we want to change everything thats purple to magenta, then we still keep the name purple, unless we refactor this everywhere... Ideally I would like to know what the purpose is of every color?

alimpens: The colors don't have a predefined purpose, this changes for each application. We could name them color1..color7. The downside of this is that it's a bit more difficult to predict which color you're going to get, but the colors would be less hard coded.

chrisvanmook: Hmm okay... In which case do you use these supplement colors anyway?

alimpens: I mostly use them for data visualisations, where the current set of primary and secondary colors are insufficient to indicate a multitude of entities (for example a line chart with several lines, a bar chart with a couple of bar groups, etc.). I think the general use case would be any situation where you would need to distinguish between several distinct groups of elements.

chrisvanmook: Alright, I'm okay with the color names then... Can't really figure out a better way. @jonkoops @royderks any opinions?

jonkoops: I would also propose to have these color use names that do not reflect their actual color. My question is whether these colors are actually part of the design system, because if they are not I would propose to keep them in the project itself and not in styled components.

alimpens: Yes, the colors are part of the [design system](https://designsystem.amsterdam.nl/7awj1hc9f/p/63dbfc-colours/t/e4d8).

jonkoops: In that case it's fine by me, if we don't like the terminology used in the design system that is something we can discuss in our bi-weekly meeting.

jonkoops: In that case it's fine by me, if we don't like the terminology used in the design system we can discuss in in our bi-weekly meetings.

PR #574, opened 2020-05-13
Icon color prop now string

jasperswart:
in the Icon component refactored color prop to be string

### DIFF
--- a/packages/asc-ui/src/components/Icon/Icon.stories.tsx
+++ b/packages/asc-ui/src/components/Icon/Icon.stories.tsx
@@ -1,5 +1,6 @@
 import { Location } from '@datapunt/asc-assets'
 import React from 'react'
+import { themeColor } from '../../utils'
 import Icon from './Icon'
 
 export default {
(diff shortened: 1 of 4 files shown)


packages/asc-ui/src/components/Icon/IconStyle.ts
```diff
@@ -43,7 +41,11 @@ const IconStyle = styled.span<Props>`
     `}
   }
 
-  ${({ color, theme }) => svgFill(color)({ theme })};
+  ${({ color }) => css`
```
chrisvanmook, line 44: Let's use the `svgFill` util here

jonkoops: @chrisvanmook `svgFill` should also be refactored actually so that it takes a color directly, or is there an easy fix here?

packages/asc-ui/src/components/Icon/IconStyle.ts
```diff
@@ -1,10 +1,8 @@
 import styled, { css } from 'styled-components'
```
jonkoops, line 1: Could you also add the breaking changes to the changelog @jasperswart?

packages/asc-ui/src/components/Icon/IconStyle.ts
chrisvanmook: Yes it needs to be refactored. Out of scope for this ticket, I'll create an issue. @jasperswart could you make the css like so:
```
& svg {
            circle,
            rect,
            polygon,
            path {
              fill: ${color};
            }
}
```

jasperswart: the svgFill is used 46 times in a lot of differtent components. if we break svgFill then those will break as well.

packages/asc-ui/src/components/Icon/IconStyle.ts
jasperswart: yes I will do that

packages/asc-ui/src/components/Icon/IconStyle.ts
jasperswart: or maybe add: svgFillColor util for this variant.

jasperswart: @chrisvanmook I have changed the css as you asked

jonkoops: Yeah, let's use the example CSS that @chrisvanmook showed. We can use the other ticket to refactor it.

packages/asc-ui/src/components/Icon/Icon.stories.tsx
```diff
@@ -109,46 +110,22 @@ export const Rotation = () => (
...
-        <Location />
-      </Icon>
-      error
-    </div>
-    <div style={{ paddingTop: 20 }}>
-      <Icon size={40} color="default">
-        <Location />
-      </Icon>
-      default
+      hotpink (any color)
```
jonkoops, line 128: Could we also add some examples here using hexadecimal and RGBA colors? Just to show people that they can pass any CSS color value into it.

packages/asc-ui/src/components/Icon/IconStyle.ts
jonkoops: Looks good!

packages/asc-ui/src/components/Icon/Icon.stories.tsx
jonkoops: This is more of a pet peeve, let's leave it until we stabilize the icons.

jonkoops:
@jasperswart this looks good to me, could you squash your changes and merge?

Issue #519, opened 2020-04-30
Remove the label/error message functionality from the Select and TextField component


robinpiets:
- Remove the label/error message functionality from the `Select` component.
- Remove the label/error message functionality from the `TextField` component.
- Replace their stories with the new `ErrorMessage` and convert from `tsx` to `mdx`

See [this comment](https://github.com/Amsterdam/amsterdam-styled-components/pull/490#issuecomment-621445365) by Jon

PR #575, opened 2020-05-13
Removed the errorMessage prop in Select

robinpiets:
See #519


packages/asc-ui/src/components/Select/Select.stories.mdx
```diff
@@ -0,0 +1,142 @@
+import { createRef, useEffect } from 'react'
+import { withKnobs, boolean, text } from '@storybook/addon-knobs'
```
chrisvanmook, line 2: These aren't used, but you can make a useful story with these! Checkout `Accordion.stories.mdx`

packages/asc-ui/src/components/Select/Select.tsx
```diff
@@ -12,15 +12,20 @@ const Select = React.forwardRef<
...
+      error,
       errorMessage,
       labelStyle,
-      errorStyle,
       ...otherProps
     },
     ref,
   ) => {
-    const error = errorProp || errorMessage
+    if (errorMessage) {
```
chrisvanmook, line 22: I think we can either show this warning, and still keep the `FormLabelStyle` (and don't mark it in the changelog as breaking, but deprecated.) OR we can remove this and consider it a breaking change

jonkoops: Let's remove it entirely.

robinpiets: 👍

packages/asc-ui/src/components/Select/Select.stories.mdx
robinpiets: Cool! I removed them for now.

jonkoops:
@robinpiets LGTM, can you squash your commits before merging?

robinpiets:
> @robinpiets LGTM, can you squash your commits before merging?

Done!

PR #615, opened 2020-05-26
Handle large text overflow in Select component

### LINKED ISSUE #614: Click on chevron doesn't work in Select component
**Describe the bug**
Click on chevron doesn't work in Select component
Also a long text will overlap the chevron

**To Reproduce**
1. Go to the select page in storybook
2. Click on the chevron icon
3. Notice that the select doesn't open

aditudorache:
This PR adds the capability of long texts in the Select component. When the text is too long, the overflow will be shown with ellipsis.

Closes #614

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,10 @@ Prefix the change with one of these keywords:
 ## Unreleased
 
 - Changed: updated the styling of `Alert` conform to design system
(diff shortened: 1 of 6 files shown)


jonkoops:
@robinpiets Unless I am mistaken, this looks to be an existing issue that is unrelated to the changes in this PR. Could you create a new issue for this?

PR #629, opened 2020-05-28
Added deprecatedWarning + Deprecated the color prop in Typography + Paragraph

robinpiets:
- Added a general deprecated console warning avoided by lint
- Deprecated the `color` prop in `Typography`
- Fixed test warnings from `MenuFlyOut/MenuFlyOut.test.tsx`

### DIFF
--- a/packages/asc-ui/src/components/ContextMenu/ContextMenu.tsx
+++ b/packages/asc-ui/src/components/ContextMenu/ContextMenu.tsx
@@ -7,6 +7,7 @@ import ContextMenuWrapperStyle from './ContextMenuWrapperStyle'
 import ContextMenuItem from './ContextMenuItem'
 import useFocusWithArrows from '../../utils/hooks/useFocusWithArrows'
 import useDetectTouchscreen from '../../utils/hooks/useDetectTouchScreen'
+import { deprecatedWarning } from '../../utils'
 
 export type Props = {
   position?: Position
@@ -36,9 +37,8 @@ const ContextMenu: React.FC<
     React.Children.toArray(children).forEach((child) => {
       // @ts-ignore
       if (child && child.type !== ContextMenuItem) {
-        // eslint-disable-next-line no-console
-        console.warn(
-          'Warning: you are rendering a different component type in <ContextMenu /> other than <ContextMenuItem />.',
+        deprecatedWarning(
+          'You are rendering a different component type in <ContextMenu /> other than <ContextMenuItem />.',
         )
       }
     })
--- a/packages/asc-ui/src/components/Typography/Typography.tsx
+++ b/packages/asc-ui/src/components/Typography/Typography.tsx
@@ -2,11 +2,17 @@ import React from 'react'
 import TypographyStyle, {
   Props as TypographyStyleProps,
 } from './TypographyStyle'
+import { deprecatedWarning } from '../../utils'
 
 export type Props = TypographyStyleProps & React.HTMLAttributes<HTMLElement>
 
-const Typography: React.FC<Props> = ({ children, ...otherProps }) => (
-  <TypographyStyle {...otherProps}>{children}</TypographyStyle>
-)
+const Typography: React.FC<Props> = ({ children, ...otherProps }) => {
+  if (otherProps.color) {
+    deprecatedWarning(
+      'You are using the deprecated `color` prop in the `Typography` component. Please use your own custom style rules',
+    )
+  }
+  return <TypographyStyle {...otherProps}>{children}</TypographyStyle>
+}
 
 export default Typography
--- a/packages/asc-ui/src/components/Typography/TypographyStyle.ts
+++ b/packages/asc-ui/src/components/Typography/TypographyStyle.ts
@@ -5,12 +5,16 @@ import { Theme } from '../../types'
 export type Props = {
   gutterBottom?: number
   element?: Variant
-  color?: Theme.ColorType
   fontSize?: number
   styleAs?: keyof Theme.TypographyElements
   as?: any
   forwardedAs?: any
   strong?: boolean
+  /**
+   * Deprecated: Use your own custom style rules
+   * @deprecated
+   */
+  color?: Theme.ColorType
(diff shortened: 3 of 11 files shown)


packages/asc-ui/src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -36,8 +37,7 @@ const ContextMenu: React.FC<
     React.Children.toArray(children).forEach((child) => {
       // @ts-ignore
       if (child && child.type !== ContextMenuItem) {
-        // eslint-disable-next-line no-console
-        console.warn(
+        deprecatedWarning(
           'Warning: you are rendering a different component type in <ContextMenu /> other than <ContextMenuItem />.',
```
jonkoops, line 41: I think we can drop the `Warning:` part of this message, as we're already using `console.warn` which should give enough formatting.

packages/asc-ui/src/components/Typography/Typography.tsx
```diff
@@ -2,11 +2,17 @@ import React from 'react'
...
 
 export type Props = TypographyStyleProps & React.HTMLAttributes<HTMLElement>
 
-const Typography: React.FC<Props> = ({ children, ...otherProps }) => (
-  <TypographyStyle {...otherProps}>{children}</TypographyStyle>
-)
+const Typography: React.FC<Props> = ({ children, ...otherProps }) => {
+  if (otherProps.color) {
+    deprecatedWarning(
+      'Warning: you are using the deprecated `color` prop in the `Typography` component. Please use your own custom style rules',
```
jonkoops, line 12: Same here we can drop the `Warning:` prefix

packages/asc-ui/src/components/Typography/TypographyStyle.ts
```diff
@@ -36,11 +40,11 @@ export default styled.p<Props>`
...
     `}
-  ${({ color }) =>
-    color &&
-    css`
-      color: ${themeColor(color)};
-    `}
+    ${({ color }) =>
+      color &&
+      css`
+        color: ${color};
```
jonkoops, line 46: Would this not break the existing functionality? It seems to me that it should still work the same as before, just simply being deprecated.

packages/asc-ui/src/types/Theme.ts
```diff
@@ -103,7 +103,7 @@ export namespace Theme {
       SupportPaletteInterface {}
 
   export type TypographyElementStyle = {
-    color: CSSProp
+    color: CSSProp // This will be deprecated in the future
```
jonkoops, line 106: Should this not be deprecated now? Best to add an `@deprecated` annotation here so the IDE will pick it up.

packages/asc-ui/src/utils/deprecatedWarning.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * A general deprecated console warning avoided by lint
+ */
+
+// eslint-disable-next-line import/prefer-default-export
```
jonkoops, line 5: We can use a default export here since it's a single function file.

packages/asc-ui/src/utils/deprecatedWarning.ts
```diff
@@ -0,0 +1,9 @@
...
+export const deprecatedWarning = (message: String) => {
+  // eslint-disable-next-line
```
jonkoops, line 7: Let's only disable the specific rule that is problematic here.

packages/asc-ui/src/components/ContextMenu/ContextMenu.tsx
chrisvanmook: Also, this warning is not about something that's deprecated...

packages/asc-ui/src/components/Typography/TypographyStyle.ts
robinpiets: Makes sense. I was doing two things at the same time.. I will keep the original and make it deprecated

packages/asc-ui/src/types/Theme.ts
robinpiets: Will do

packages/asc-ui/src/utils/deprecatedWarning.ts
robinpiets: 👍

packages/asc-ui/src/components/Typography/TypographyStyle.ts
```diff
@@ -5,12 +5,16 @@ import { Theme } from '../../types'
...
-  color?: Theme.ColorType
   fontSize?: number
   styleAs?: keyof Theme.TypographyElements
   as?: any
   forwardedAs?: any
   strong?: boolean
+  /**
+   * Deprecated: Use your own custom style rules
+   * @deprecated
+   */
```
jonkoops, line 16: The text you put after `@deprecated` will be used by the IDE. So I would change this to:

```suggestion
  /**
   * @deprecated Use your own custom style rules
   */
```

packages/asc-ui/src/types/Theme.ts
```diff
@@ -103,12 +103,16 @@ export namespace Theme {
...
     fontWeight: 400 | 500 | 700 | 'inherit'
     fontSize: CSSProp
     lineHeight: number | CSSProp
     letterSpacing: CSSProp
     marginBottom: CSSProp
+    /**
+     * Deprecated: Use your own custom style rules
+     * @deprecated
+     */
+    color: CSSProp
```
jonkoops, line 115: Same here

packages/asc-ui/src/components/Typography/TypographyStyle.ts
robinpiets: Sure!

packages/asc-ui/src/types/Theme.ts
robinpiets: Changed this in `ButtonStyle` as well

PR #633, opened 2020-06-03
Replaced onDarkBackground in Link with darkBackground

robinpiets:


packages/asc-ui/src/components/TopTaskLink/TopTaskLinkStyle.ts
```diff
@@ -20,7 +20,7 @@ export const TopTaskTitle = styled(Paragraph)`
 `
 
 export default styled.a<{
-  onDarkBackground?: boolean
+  hasDarkBackground?: boolean
```
jonkoops, line 23: I don't think we can simply rename this prop as we need to still support the deprecated variant.

packages/asc-ui/src/components/Alert/Alert.stories.mdx
```diff
@@ -92,14 +92,14 @@ export const Decorator = styled.div`
           Accusamus ad illo molestiae vel. Deleniti excepturi iusto neque
           quisquam!
         </Paragraph>
-        <Link onDarkBackground href="/" variant="with-chevron">
+        <Link href="/" variant="with-chevron" hasDarkBackground>
```
jonkoops, line 95: Perhaps we can just shorten this to `darkBackground`?

packages/asc-ui/src/components/TopTaskLink/TopTaskLinkStyle.ts
robinpiets: Good catch.

Fixed all improvements 👍

PR #771, opened 2020-07-08
Use createEvent utility to support older browsers

jonkoops:


packages/asc-ui/src/utils/createEvent.ts
```diff
@@ -0,0 +1,11 @@
...
+  if (typeof Event === 'function') {
+    return new Event(eventName, initOptions)
+  } 
+
+  // Older browsers like Internet Explorer do not support the 'Event' constructor.
+  const event = document.createEvent('Event');
+  event.initEvent(eventName, initOptions?.bubbles, initOptions?.cancelable);
+  
+  return event
+}
```
royderks, line 11: Linting issue here

jonkoops: Yes, fixed it :)

PR #783, opened 2020-07-10
Fix link to fonts documentation

marin-amsterdam:
Without the starting `/` it is a relative link and links to a page that does not exist (`/docs/packages/asc-assets/README.md#Fonts`).


jonkoops:
Thanks for the contribution @marin-amsterdam!

PR #833, opened 2020-07-20
Added rowProps to FooterBottom

robinpiets:
For me it's very strange that `FooterBottom` has a `Row` as parent, while `FooterTop` does not. In the old Footer component I used to wrap the bottom part in my own Row/Column, but now I clearly cannot. So I've added a `rowProps` prop to be able to pass my own props, but maybe it's better to remove the `Row` in `FooterBottom` completely?

### DIFF
--- a/packages/asc-ui/src/components/Footer/FooterBottom.tsx
+++ b/packages/asc-ui/src/components/Footer/FooterBottom.tsx
@@ -3,6 +3,10 @@ import styled from 'styled-components'
 import { themeSpacing } from '../../utils'
 import { Row } from '../Grid'
 
+type Props = {
+  rowProps: object
+}
+
 const FooterBottomStyle = styled.div`
   display: flex;
   flex-direction: row;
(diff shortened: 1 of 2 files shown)


packages/asc-ui/src/components/Footer/FooterBottom.tsx
```diff
@@ -3,6 +3,10 @@ import styled from 'styled-components'
 import { themeSpacing } from '../../utils'
 import { Row } from '../Grid'
 
+type Props = {
+  rowProps: object
```
jonkoops, line 7: There should be a better type that we can use here. Does the row itself not have some props?

jonkoops:
Yeah, perhaps it's best to remove this functionality. What do you think @chrisvanmook?

packages/asc-ui/src/components/Footer/FooterBottom.tsx
robinpiets: Row has plenty of props. I just wanted to be able to pass my own `style` prop to pass my own styles. I don't really need those other props, although they can be helpful for someone else.
```
children: React.ReactNode
  className?: string
  debug?: boolean
  debugColor?: string
  halign?: FlexJustify
  hasMargin?: boolean
  hasMaxWidth?: boolean
  valign?: TypeFlexPosition
```

NB: I'm fine with closing this PR because I've found another solution in my own project, even though I still find it strange that FooterBottom has a Row as parent.

ping: @chrisvanmook

robinpiets:
I've solved this in my own project, so I don't need this solution anymore

PR #834, opened 2020-07-20
Minor styling fix on Button in FooterSection

robinpiets:
### Fixed the alignment of the Button compared to the children

---
#### From: 
image

---
#### To:
image


packages/asc-ui/src/components/Footer/FooterSectionStyle.ts
```diff
@@ -25,6 +25,9 @@ const StyledFooterHeading = styled(FooterHeading)`
 `
 
 const StyledButton = styled(Button)`
+  margin-left: -${themeSpacing(2)};
```
chrisvanmook, line 28: Perhaps a small explanation why?

PR #943, opened 2020-08-05
Removed custom focus styles, now use browser native focus

chrisvanmook:


packages/asc-ui/src/components/Button/ButtonStyle.ts
```diff
@@ -2,7 +2,6 @@ import { darken, readableColor, size, transitions } from 'polished'
 import styled, { css } from 'styled-components'
```
jonkoops, line 2: There are still two instances in this file where the `outline` is set which might interfere with the browser styling.

packages/asc-ui/src/components/FilterTag/FilterTagStyle.ts
```diff
@@ -1,37 +1,29 @@
...
 
-    ${IconStyle} {
-      margin: 0 0 0 ${themeSpacing(2)};
-      vertical-align: middle;
-    }
-  `
-}
-
-export default styled(TypographyStyle)<Props>`
-  ${FilterTagStyleCSS}
```
jonkoops, line 36: I have no idea why this was like this, but good that you re-factored it :+1:

packages/asc-ui/src/components/Checkbox/CheckboxStyle.ts
```diff
@@ -1,6 +1,5 @@
 import styled, { css } from 'styled-components'
```
jonkoops, line 1: There is still some styling present that modifies the `outline` in this file.

packages/asc-ui/src/components/ShowMoreShowLess/ShowMoreShowLessStyle.ts
```diff
@@ -5,6 +5,7 @@ export type ShowMoreShowLessProps = {
   maxHeight?: string
   showMoreText?: string
   showLessText?: string
+  gradientBackground?: string
```
jonkoops, line 8: This seems an unrelated change, I think it should be removed.

packages/asc-ui/src/index.ts
```diff
@@ -115,12 +115,6 @@ export {
   svgFill,
   themeSpacing,
 } from './utils'
-export {
-  fillWhenFocused,
-  focusFill,
-  focusOutline,
-  outlineWhenFocused,
-} from './utils/focus'
```
jonkoops, line 123: :fire: :fire: :fire:

stories/src/ui/ShowMoreShowLess.stories.mdx
```diff
@@ -1,6 +1,6 @@
 import styled from 'styled-components'
```
jonkoops, line 1: Same for this change, seems unrelated.

packages/asc-ui/src/components/Button/ButtonStyle.ts
chrisvanmook: only for hover and disabled state. shouldn't interfere I guess, right?

packages/asc-ui/src/components/Checkbox/CheckboxStyle.ts
chrisvanmook: We need the outline rules here as well.

packages/asc-ui/src/components/ShowMoreShowLess/ShowMoreShowLessStyle.ts
chrisvanmook: Will open a separate PR for this. excuse my laziness 😉

packages/asc-ui/src/components/Button/ButtonStyle.ts
jonkoops: Yeah, should be ok. Just double checking.

packages/asc-ui/src/components/Checkbox/CheckboxStyle.ts
jonkoops: :+1:

PR #975, opened 2020-08-07
Add Tabs component

jonkoops:
See the reference from the Design System: https://designsystem.amsterdam.nl/7awj1hc9f/p/15737c-tab-navigation

### DIFF
--- a/packages/asc-ui/src/components/Tabs/Tab.tsx
+++ b/packages/asc-ui/src/components/Tabs/Tab.tsx
@@ -1,4 +1,4 @@
-import React, { HTMLAttributes } from 'react'
+import React, { HTMLAttributes, ReactNode } from 'react'
 
 // Some props are omitted since they cannot be overwritten.
 type OmittedProps = 'role' | 'aria-controls' | 'aria-selected' | 'tabIndex'
@@ -6,7 +6,7 @@ type OmittedProps = 'role' | 'aria-controls' | 'aria-selected' | 'tabIndex'
 export interface TabProps
   extends Omit<HTMLAttributes<HTMLButtonElement>, OmittedProps> {
   id: string
-  label: string
+  label: ReactNode
 }
 
 export const Tab: React.FC<TabProps> = () => {
--- a/packages/asc-ui/src/components/Tabs/Tab.tsx
+++ b/packages/asc-ui/src/components/Tabs/Tab.tsx
@@ -1,0 +1,14 @@
+import React, { HTMLAttributes } from 'react'
+
+// Some props are omitted since they cannot be overwritten.
+type OmittedProps = 'role' | 'aria-controls' | 'aria-selected' | 'tabIndex'
+
+export interface TabProps
+  extends Omit<HTMLAttributes<HTMLButtonElement>, OmittedProps> {
+  id: string
+  label: string
+}
+
+export const Tab: React.FC<TabProps> = () => {
+  return null
+}
--- a/packages/asc-ui/src/components/Tabs/TabButton.tsx
+++ b/packages/asc-ui/src/components/Tabs/TabButton.tsx
@@ -5,10 +5,10 @@ const TabButton = styled.button.attrs({
   role: 'tab',
 })<{ isSelected?: boolean }>`
   background-color: transparent;
-  padding: ${themeSpacing(3, 4)};
+  padding: ${themeSpacing(3, 0)};
   color: ${themeColor('tint', 'level7')};
   border: none;
-  margin-right: ${themeSpacing(8)};
+  margin-right: ${themeSpacing(7)};
   transition: box-shadow 0.1s ease-in-out, color 0.1s ease-in-out;
   font-size: 18px;
   font-weight: 700;
--- a/packages/asc-ui/src/components/Tabs/TabButton.tsx
+++ b/packages/asc-ui/src/components/Tabs/TabButton.tsx
@@ -1,0 +1,33 @@
+import styled, { css } from 'styled-components'
+import { themeColor, themeSpacing } from '../../utils'
+
+const TabButton = styled.button.attrs({
+  role: 'tab',
+})<{ isSelected?: boolean }>`
+  background-color: transparent;
+  padding: ${themeSpacing(3, 4)};
+  color: ${themeColor('tint', 'level7')};
+  border: none;
+  margin-right: ${themeSpacing(8)};
+  transition: box-shadow 0.1s ease-in-out, color 0.1s ease-in-out;
+  font-size: 18px;
+  font-weight: 700;
+  cursor: pointer;
+
+  &:hover {
+    box-shadow: 0 -6px inset ${themeColor('tint', 'level7')};
+  }
+
+  ${({ isSelected }) =>
+    isSelected &&
+    css`
+      color: ${themeColor('secondary', 'main')};
+      box-shadow: 0 -6px inset ${themeColor('secondary', 'main')};
+
+      &:hover {
+        box-shadow: 0 -6px inset ${themeColor('secondary', 'main')};
+      }
+    `}
+`
+
+export default TabButton
--- a/packages/asc-ui/src/components/Tabs/TabList.tsx
+++ b/packages/asc-ui/src/components/Tabs/TabList.tsx
@@ -1,0 +1,31 @@
+import styled from 'styled-components'
+import { themeColor, themeSpacing } from '../../utils'
+
+const TabList = styled.div.attrs({
+  role: 'tablist',
+})`
+  min-width: 100%;
+  box-shadow: inset 0 -2px 0 0 ${themeColor('tint', 'level2')};
+  white-space: nowrap;
+  overflow-x: scroll;
+  scrollbar-width: none; /* Firefox */
+  -ms-overflow-style: none; /* Internet Explorer 10+ */
+  mask: linear-gradient(
+      to right,
+      black calc(100% - ${themeSpacing(8)}),
+      transparent
+    ),
+    linear-gradient(
+      to bottom,
+      transparent calc(100% - 6px),
+      black calc(100% - 6px)
+    );
+
+  &::-webkit-scrollbar {
+    /* WebKit */
+    width: 0;
+    height: 0;
+  }
+`
+
+export default TabList
--- a/packages/asc-ui/src/components/Tabs/Tabs.tsx
+++ b/packages/asc-ui/src/components/Tabs/Tabs.tsx
@@ -49,14 +49,16 @@ export const Tabs: React.FC<TabsProps> = ({ label, children }) => {
     <>
       <TabList aria-label={label} onKeyDown={keyDown} ref={tabListRef}>
         {children.map(({ props }) => {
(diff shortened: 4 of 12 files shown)


packages/asc-ui/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,138 @@
...
+          const tabId = formatTabId(props.id)
+          const panelId = formatPanelId(props.id)
+
+          return (
+            <TabButton
+              key={props.id}
+              id={tabId}
+              aria-controls={panelId}
+              aria-selected={isSelected}
+              tabIndex={isFocusable ? 0 : -1}
```
alimpens, line 111: tabindex shouldn't change on focus, only on selection. See implementation [here](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-2/tabs.html). The problem with changing tabindex on focus is that when you leave the tab list and return, the focus also returns to the last focussed tab, not the selected tab. This doesn't match with _When focus moves into the tab list, places focus on the active tab element_ from that same link.

packages/asc-ui/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const focusPreviousTab = () => focusTowards(FocusDirection.Backward)
+
+  function handleKeyDown(event: React.KeyboardEvent<HTMLDivElement>) {
+    if (event.key === 'ArrowRight') {
+      focusNextTab()
+    }
+
+    if (event.key === 'ArrowLeft') {
+      focusPreviousTab()
+    }
```
alimpens, line 93: Do we also want to provide Home- and End-button functionality? You can find an example of that [here](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-2/tabs.html).

chrisvanmook: We actually have a hook for this that we can use here: `useFocusWithArrows`, which includes the Home and End buttons

packages/asc-ui/src/utils/hooks/useFocusWithArrows.tsx
```diff
@@ -8,12 +8,16 @@ import { FOCUSABLE_ELEMENTS } from './constants'
  * @param ref Component ref
  * @param rotating Jump to first item from last or vice versa
  * @param directChildrenOnly Useful if you don't want to focus on other focussable elements inside the child components of the ref
+ * @param horizontally In case you need to navigate horizontally, using left / right arrow buttons
  */
 const useFocusWithArrows = (
   ref: React.RefObject<HTMLElement>,
   rotating = false,
   directChildrenOnly = false,
+  horizontally = false,
```
chrisvanmook, line 17: Probably need to refactor this to accept an object with options, now we need to pass 4 arguments

jonkoops: Yeah, agreed this is a tad confusing right now.

stories/src/ui/Tabs.stories.mdx
```diff
@@ -0,0 +1,28 @@
...
+      </Tab>
+      <Tab id="two" label="Second">
+        Contents of the second tab.
+      </Tab>
+      <Tab id="three" label="Last">
+        Contents of the last tab.
+      </Tab>
+    </Tabs>
+  </Story>
+</Preview>
```
chrisvanmook, line 28: I saw that it also works well on mobile with the horizontal scroll, nice! Should we add a mobile story here as well just for convenience? You can use a storybook addon for this

packages/asc-ui/src/components/Tabs/TabButtonStyle.ts
```diff
@@ -0,0 +1,35 @@
...
+type Props = {
+  isSelected?: boolean
+}
+
+const TabButtonStyle = styled.button<Props>`
+  background-color: transparent;
+  padding: 12px 15px;
+  color: ${themeColor('tint', 'level7')};
+  border: none;
+  margin-right: 30px;
```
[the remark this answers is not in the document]
jonkoops: Fixed

packages/asc-ui/src/components/Tabs/TabButtonStyle.ts
```diff
@@ -0,0 +1,35 @@
+import styled, { css } from 'styled-components'
+import { themeColor } from '../../utils'
+
+type Props = {
+  isSelected?: boolean
+}
+
+const TabButtonStyle = styled.button<Props>`
+  background-color: transparent;
+  padding: 12px 15px;
```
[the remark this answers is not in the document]
jonkoops: Fixed

alimpens (approved):

packages/asc-ui/src/components/Tabs/TabList.tsx
```diff
@@ -1,45 +1,26 @@
...
-  useImperativeHandle,
-  useRef,
-} from 'react'
-import TabListStyle, { TabListWrapper, TabListFade } from './TabListStyle'
+import styled from 'styled-components'
+import { themeColor, themeSpacing } from '../../utils'
 
-export default forwardRef<HTMLDivElement, HTMLAttributes<HTMLDivElement>>(
-  ({ children, ...otherProps }, externalRef) => {
-    const [canScroll, setCanScroll] = useState<boolean>(false)
```
alimpens: Without checking if TabList can scroll you always have a small fade on the right, which you can see over the inset box shadow. It's quite small, but it can imply for some users that TabList is scrollable when it isn't. Should we keep canScroll here?

packages/asc-ui/src/components/Tabs/Tabs.tsx
```diff
@@ -33,7 +33,7 @@ function formatPanelId(id: string) {
   return `panel-${id}`
 }
 
-export const Tabs: React.FC<Tabs> = ({ label, children, ...otherProps }) => {
```
alimpens: Why did you remove `otherProps` here? The idea was that the user can set props (such as `style`) on the `Tabs` component, in order to style `TabList `for example.

packages/asc-ui/src/components/Tabs/Tabs.tsx
```diff
@@ -43,38 +43,28 @@ export const Tabs: React.FC<Tabs> = ({ label, children, ...otherProps }) => {
...
   return (
     <>
-      <TabList
-        aria-label={label}
-        onKeyDown={keyDown}
-        ref={tabListRef}
-        {...otherProps}
-      >
-        {children.map(
-          ({ props: { id, label: childLabel, ...otherChildProps } }) => {
```
alimpens: Why did you refactor this? Destructuring the props is more readable, in my opinion. And why did you remove `otherChildProps`? The idea was to enable the user to set props (such as `style`) on `Tab`, in order to style `TabButton`.

packages/asc-ui/src/components/Tabs/Tabs.tsx
alimpens: This behaviour seems to be broken again, the focus seems to return to the last focussed tab, not the selected tab. I can't quite figure out why this changed, though :)

alimpens (commented):
This commit depends on the answer to [this comment](https://github.com/Amsterdam/amsterdam-styled-components/pull/975/commits/f0679515bbab4d15ad7e567e65a2d706e5b30277#r473926714), I think.

timvanoostrom:
Also, I've compared the design with the one referenced in designsystem. 
Screen Shot 2020-08-24 at 09 21 44
Screen Shot 2020-08-24 at 09 22 43

A couple of things stand out:

- there is no padding in the tab buttons in the design reference
- the font seems to be larger and possibly not the same weight

timvanoostrom:
I see the font specs are written out so maybe the design or the specs are wrong. I guess some clarification is needed.

packages/asc-ui/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,96 @@
...
+const TabPanel = styled.div.attrs({
+  role: 'tabpanel',
+  tabIndex: 0,
+})``
+
+export interface TabsProps {
+  /**
+   * A descriptive label for the tabs, required for accessibility.
+   */
+  label: string
```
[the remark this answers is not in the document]
alimpens: This `label` is used to set `aria-label`, you can't use ReactNode there. I assume you mean the `label` in Tab.tsx?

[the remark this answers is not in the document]
timvanoostrom: Ah, It's actually used for both. So maybe we should make distinction. 

```
label: ReactNode
ariaLabel: string;
```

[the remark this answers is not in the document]
timvanoostrom: Ah I see, then it applies to: https://github.com/Amsterdam/amsterdam-styled-components/pull/975/files/5b124f88b20580a137a9b0ac1c3a364e9b46b6d2#diff-c14a52293bc086d98059a5983e3571a8R9

packages/asc-ui/src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { HTMLAttributes } from 'react'
+
+// Some props are omitted since they cannot be overwritten.
+type OmittedProps = 'role' | 'aria-controls' | 'aria-selected' | 'tabIndex'
+
+export interface TabProps
+  extends Omit<HTMLAttributes<HTMLButtonElement>, OmittedProps> {
+  id: string
+  label: string
```
timvanoostrom, line 9: You could make this `ReactNode` so that some other elements than `string` can be passed. For example an `<Icon />`.

packages/asc-ui/src/components/Tabs/Tabs.tsx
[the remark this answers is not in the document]
timvanoostrom: Does a component like this need an `aria-label` at all ? It's a `role="tablist"` right?

https://www.marcozehe.de/advanced-aria-tip-1-tabs-in-web-apps/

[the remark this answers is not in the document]
timvanoostrom: Nah, W3 says `Provides a label that describes the purpose of the set of tabs.` so all good.

timvanoostrom:
LGTM!

chrisvanmook (approved):

packages/asc-ui/src/components/Tabs/TabButton.tsx
```diff
@@ -5,10 +5,10 @@ const TabButton = styled.button.attrs({
   role: 'tab',
 })<{ isSelected?: boolean }>`
   background-color: transparent;
-  padding: ${themeSpacing(3, 4)};
+  padding: ${themeSpacing(3, 0)};
   color: ${themeColor('tint', 'level7')};
   border: none;
-  margin-right: ${themeSpacing(8)};
+  margin-right: ${themeSpacing(7)};
```
jonkoops, line 11: @alimpens Be aware that the fade should also be changed if this is changed.

PR #1122, opened 2020-09-01
Feature/assets icons

gilleswittenberg:
- removed executable rights from some existing .svg files
- added 3 missing SVG icons

## Questions / todos
- I see more missing icons that are shown on https://designsystem.amsterdam.nl/7awj1hc9f/p/75928b-icons. Do we want to add them all? Or maybe nicer: write a little script to pull + unpack them all?
- yarn.lock has diffs locally. Should these be added to the PR?
- The processing of the .svg files was not run as a pre-push hook. Is this too heavy to run for all pushes? Can we maybe run it only on new / changed icons?
- Should this be added to CHANGELOG.md?


gilleswittenberg:
Assume the failing checks are not caused by the changes in this PR. Let me know if they are.

jonkoops:
> I see more missing icons that are shown on https://designsystem.amsterdam.nl/7awj1hc9f/p/75928b-icons. Do we want to add them all? Or maybe nicer: write a little script to pull + unpack them all?

This eventually something I would like to end up with as well, preferably the pipeline for getting icons from the design system would be automated. This seems out of scope for this PR but feel free to create an issue for it.

> yarn.lock has diffs locally. Should these be added to the PR?

Ideally yes, but it might cause merge conflicts later on. This happens because dependabot updates the lockfile, but it only touches the specific version of the dependency it's updated. This means that packages that are locked but no longer needed are not removed, once you run yarn install Yarn will attempt to get rid of these.

Basically don't worry about it, it cannot do much harm. We will periodically update the lockfile by hand. If you want to prevent yarn from changing your local lockfile you can install dependencies with `yarn install --frozen-lockfile`.

> Should this be added to CHANGELOG.md?

Yes please :)

jonkoops (changes requested):
Could you add an entry to the changelog as well?

gilleswittenberg:
Ok. Thanks. Will finish up next week.

jonkoops (approved):

chrisvanmook (approved):

PR #1157, opened 2020-09-14
Added new variants to `Alert`

robinpiets:
### This PR is all about adding the two new variants for the `Alert` component: `valid` and `focus`. As discussed with DT. 

<hr>

### I named them according to their color-name

<hr>

### DIFF
--- a/packages/asc-ui/src/components/Alert/AlertStyle.ts
+++ b/packages/asc-ui/src/components/Alert/AlertStyle.ts
@@ -4,7 +4,13 @@ import { svgFill, themeColor, themeSpacing } from '../../utils'
 import Button from '../Button'
 import Heading from '../Heading'
 
-export type Level = 'normal' | 'attention' | 'warning' | 'error'
+export type Level =
+  | 'normal'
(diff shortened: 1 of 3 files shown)


timvanoostrom:
LGTM

packages/asc-ui/src/components/Alert/AlertStyle.ts
```diff
@@ -4,7 +4,13 @@ import { svgFill, themeColor, themeSpacing } from '../../utils'
...
 import Heading from '../Heading'
 
-export type Level = 'normal' | 'attention' | 'warning' | 'error'
+export type Level =
+  | 'normal'
+  | 'attention'
+  | 'warning'
+  | 'error'
+  | 'valid'
+  | 'focus'
```
jonkoops, line 13: I think I'd prefer to use terminology here that represents the state rather than the color. Besides that I think we need to discuss which variants should be supported because this is getting kinda out of hand.

robinpiets:
Closing this PR because I created a new PR #1295

PR #1160, opened 2020-09-16
(Chore) Fix double onChange call in Select component

janjaap:
This PR fixes an issue where the `onChange` callback handler on the `Select` component was called twice; once on component instantiation and once on actual change of the underlying element's value.

Assertions have been added to validate the changes. 

Note that the `cleanup()` call has been removed; cleanup after each test case is default as well as synchronous (so, when needed to call, needs to wrapped in `act()`). Also, events that alter the component's state, need to be wrapped in `act()`; has also been applied.


packages/asc-ui/src/components/Select/Select.test.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
```
jonkoops, line 1: Thanks for cleaning up this code!

packages/asc-ui/src/components/Select/Select.tsx
```diff
@@ -82,16 +80,6 @@ const Select = React.forwardRef<
...
-    useEffect(() => {
-      if (inputValue && ref.current) {
-        setValue(inputValue)
-
-        // The selectedValue label doesn't update properly if you send a new value prop to this component and the component doesn't rerender
-        ref.current.value = inputValue
-        ref.current.dispatchEvent(createEvent('change', { bubbles: true }))
-      }
-    }, [inputValue, ref])
-
```
jonkoops, line 94: Now that this code has been removed are we not introducing the original bug it attempted to fix? If possible we should add a test that changes the `value` prop to see if the overlaying label gets updated properly.

jonkoops (changes requested):

packages/asc-ui/src/components/Select/Select.tsx
janjaap: I've tested the behaviour in several browsers and it all works as expected. I'll add a test that will cover the display value of the overlaying label.

jonkoops (approved):

jonkoops:
@janjaap Code looks good, could you add an entry to the changelog?

janjaap:
> @janjaap Code looks good, could you add an entry to the changelog?

Done. And added another test that will verify that programmatically changing the value will actually update the overlay label.

Issue #1289, opened 2020-10-28
Modal component: zIndexOffset prop warning


strexx:
**Description**
When triggering the Modal, React gives a warning about a `zIndexOffset` prop that is not recognized.

ASC version used in this project: `0.21.2-alpha.22`

**Video**
![modal-bug](https://user-images.githubusercontent.com/1044742/97446982-5b3fd780-192f-11eb-8cdf-9778e84bb839.gif)

**Desktop (please complete the following information):**
 - OS: Mac Catalina
 - Browser: Chrome
 - Version 86.0.4240.11

PR #1299, opened 2020-10-29
Fixed the Modal component `zIndexOffset` prop warning 

robinpiets:
When triggering the Modal, React ~~gives~~ gave a warning about a `zIndexOffset` prop that is not recognised.

See #1289

### DIFF
--- a/packages/asc-ui/src/components/Focus/Focus.tsx
+++ b/packages/asc-ui/src/components/Focus/Focus.tsx
@@ -1,9 +1,12 @@
 /* eslint-disable jsx-a11y/no-noninteractive-tabindex */
 import React from 'react'
+import styled from 'styled-components'
 
 type Props = {
   onKeyDown?: React.EventHandler<React.KeyboardEvent>
 }
+
+const FocusStyle = styled.div``
 
 const Focus: React.FC<Props> = ({ children, onKeyDown, ...otherProps }) => {
   const myRef = React.createRef<HTMLDivElement>()
@@ -28,7 +31,7 @@ const Focus: React.FC<Props> = ({ children, onKeyDown, ...otherProps }) => {
   }, [myRef])
 
   return (
-    <div
+    <FocusStyle
       {...otherProps}
       role="presentation"
       ref={myRef}
@@ -36,7 +39,7 @@ const Focus: React.FC<Props> = ({ children, onKeyDown, ...otherProps }) => {
       tabIndex={0}
     >
       {children}
-    </div>
+    </FocusStyle>
   )
 }
 


strexx (commented):
LGTM

packages/asc-ui/src/components/Focus/Focus.tsx
```diff
@@ -1,10 +1,13 @@
 /* eslint-disable jsx-a11y/no-noninteractive-tabindex */
-import React from 'react'
+import React, { HTMLAttributes } from 'react'
+import styled from 'styled-components'
 
 type Props = {
   onKeyDown?: React.EventHandler<React.KeyboardEvent>
 }
 
+const FocusStyle = styled.div<HTMLAttributes<HTMLDivElement>>``
```
jonkoops, line 9: Do we need to pass this generic type? I think it can be inferred from `styled.div`.

```suggestion
const FocusStyle = styled.div``
```

packages/asc-ui/src/components/Focus/Focus.tsx
```diff
@@ -28,15 +31,15 @@ const Focus: React.FC<Props> = ({ children, onKeyDown, ...otherProps }) => {
   }, [myRef])
 
   return (
-    <div
+    <FocusStyle
```
jonkoops, line 34: Could you elaborate a bit on how this solves the issue?

jonkoops (changes requested):

packages/asc-ui/src/components/Focus/Focus.tsx
robinpiets: True, good point

packages/asc-ui/src/components/Focus/Focus.tsx
robinpiets: If you place a prop like `zIndexOffset` directly on a `div` element, React will trigger a warning that it doesn't recognise that attribute. However, if you pass that prop to a `styled` component, `styled-components` handles it for you and will not place that prop directly on the element.

We only needs that prop in `ModalFocus` (which is an instance of `Focus` to make `showAboveBackDrop()` work) and we don't need it as a HTML attribute.

The `BackDropStyle` that uses the `zIndexOffset` is also a `styled.div` and that prop is not added to the `div` as well. 

The Modal component is still working. To proof this you can remove the `zIndexOffset` prop from `<ModalFocus />` and the Modal will fail to render.

packages/asc-ui/src/components/Focus/Focus.tsx
robinpiets: Pushed

packages/asc-ui/src/components/Focus/Focus.tsx
jonkoops: Ok that makes sense, although it's not very clear in the code so I can see someone removing it in the future by accident.

jonkoops (approved):

PR #1321, opened 2020-11-04
Made ContextMenu respond dynamically to `open` prop.

alimpens:

- [ ] You've added or updated the README.md of the component
- [ ] You've exported the component in [index.ts](./packages/asc-ui/src/index.ts)


packages/asc-ui/src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -45,6 +45,15 @@ const ContextMenu: React.FC<
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
+  // Set state with 'open' prop if it's set
+  useEffect(() => {
+    if (!openProp) {
+      setOpen(false)
+    } else if (openProp) {
+      setOpen(true)
+    }
```
janjaap, line 54: You could suffice with `setOpen(openProp)`

jonkoops, line 54: It's best to see if the value is different before calling `setOpen` so that you don't cause any unnecessary re-renders of the component. You can also pass in the value directly instead of using a condition, for example:

```suggestion
    if (openProp !== open) {
      setOpen(openProp)
    }
```

Note that in order to do so you will have to provide a default value for `openProp`, otherwise it might end up being `undefined`.

```ts
open: openProp = false,
```

This also means that we don't have to handle falling back in the original `useState`:

```
const [open, setOpen] = useState(openProp) 
```

jonkoops (changes requested):

jonkoops (approved):

PR #1384, opened 2020-11-25
Show default browser focus style in the select component

aditudorache:
This PR:
- shows the default browser focus style in the select component
- exports the `SelectContentWrapperStyle` to ease the customization 
- sets the `className` on the `SelectWrapper` so the component can be properly styled


robinpiets (approved):
Nice change. It looks better now

PR #1387, opened 2020-11-25
Change the html container for the logo (WCAG)

aditudorache:
This PR:
- changes the html element in the logo from `h1` to `div` for wcag compliance

Note: the changelog will be updated when consent is reached

### DIFF
--- a/packages/asc-ui/src/components/Header/HeaderLogoTextStyle.ts
+++ b/packages/asc-ui/src/components/Header/HeaderLogoTextStyle.ts
@@ -1,6 +1,6 @@
 import styled from 'styled-components'
 
-const HeaderLogoTextStyle = styled.h1`
+const HeaderLogoTextStyle = styled.div`
   margin: 0;
   font-weight: 700;
   display: flex;


robinpiets:
Why is this needed and did it come out of a WCAG audit? Can we not add a function to use `as` or `forwardedAs` to overwrite the default `h1`?

packages/asc-ui/src/components/Header/HeaderLogoTextStyle.ts
```diff
@@ -1,6 +1,6 @@
 import styled from 'styled-components'
 
-const HeaderLogoTextStyle = styled.h1`
+const HeaderLogoTextStyle = styled.div`
```
janjaap, line 3: You'd probably want this to be configurable, because on a homepage, it makes sense to have it be an `h1`.

aditudorache: It make sense but not to wrap the site logo in it. This is the wcag suggestion. Besides that this control is actualy rendered in a <header> html element that makes sense.

aditudorache:
@robinpiets if you can help me with a snippet I'll consider changing this just for `SIA`. Self I could't find a way without changing the `asc`.

robinpiets:
> @robinpiets if you can help me with a snippet I'll consider changing this just for `SIA`. I could't find a way without changing the `asc`.

You need to change `asc-ui`, otherwise you cannot update nested components via `as`. How about this? https://github.com/Amsterdam/amsterdam-styled-components/pull/1388

Do you have another suggestion? @jonkoops @janjaap

aditudorache:
closed in favor of https://github.com/Amsterdam/amsterdam-styled-components/pull/1388

jonkoops:
Closed in favour of #1388

PR #1388, opened 2020-11-25
Added `headerLogoTextAs` prop to `Header` component to overwrite the default `h1` tag

robinpiets:
Demo for @aditudorache 

#### with `headerLogoTextAs="div"`:
image

#### without `headerLogoTextAs` prop:
image

### DIFF
--- a/packages/asc-ui/src/components/Header/HeaderLogoText.tsx
+++ b/packages/asc-ui/src/components/Header/HeaderLogoText.tsx
@@ -1,6 +1,6 @@
 import React, {
   AnchorHTMLAttributes,
-  ComponentType,
+  ElementType,
   FunctionComponent,
   HTMLAttributes,
 } from 'react'
@@ -14,7 +14,7 @@ export interface LogoProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
 }
 
 interface HeaderLogoTextProps extends HTMLAttributes<HTMLHeadingElement> {
-  as?: string | ComponentType
+  as?: string | ElementType
   homeLink: string
   tall?: boolean
   logo?: FunctionComponent<LogoProps>
--- a/packages/asc-ui/src/components/Header/HeaderLogoText.tsx
+++ b/packages/asc-ui/src/components/Header/HeaderLogoText.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import { StyledProps } from 'styled-components'
 import HeaderLogoTextStyle from './HeaderLogoTextStyle'
 import HeaderTitle from './HeaderTitle'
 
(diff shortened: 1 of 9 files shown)


aditudorache (approved):
Good improvement! Le't do this then!

robinpiets:
What do you think @jonkoops @janjaap @chrisvanmook ?

stories/src/ui/Header.stories.mdx
```diff
@@ -20,7 +20,12 @@ Given the `tall={false}` prop, this will make sure the header is not showing its
 <Preview>
   <Story name="Short only">
     <React.Fragment>
-      <Header tall={false} title="Some title" fullWidth={false} />
+      <Header
```
robinpiets, line 23: TODO: remove comment and create new story for this

packages/asc-ui/src/components/Header/HeaderLogoText.tsx
```diff
@@ -1,27 +1,27 @@
...
 
-const HeaderLogoText: React.FC<HeaderLogoTextProps> = ({
-  title,
-  homeLink,
-  tall,
-  logo: LogoIcon,
-  ...otherProps
-}) => (
+const HeaderLogoText: FunctionComponent<
+  HeaderLogoTextProps & StyledProps<any>
```
jonkoops, line 23: It doesn't look like `StyledProps` actually adds the `as` prop to the type definition, but rather seems to ignore it. Perhaps the `StyledComponent` type would be a better option here?

stories/src/ui/Header.stories.mdx
```diff
@@ -20,7 +20,12 @@ Given the `tall={false}` prop, this will make sure the header is not showing its
...
+        headerLogoTextAs="div" // Now the HeaderLogoText component will be a `div`
```
jonkoops, line 24: Should this be a separate story entry?

jonkoops (changes requested):

stories/src/ui/Header.stories.mdx
robinpiets: Yep. Done

packages/asc-ui/src/components/Header/HeaderLogoText.tsx
```diff
@@ -1,28 +1,33 @@
...
-  extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+export interface LogoProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
   href: string
   tall?: boolean
   title?: string
 }
 
-interface HeaderLogoTextProps extends React.HTMLAttributes<HTMLHeadingElement> {
+interface HeaderLogoTextProps extends HTMLAttributes<HTMLHeadingElement> {
+  as?: string | React.ComponentType
```
jonkoops, line 16: Can you import `ComponentType` directly from React so that we do not have to use the `React` namespace. I was thinking perhaps we could use `ElementType` instead? I believe it actually accepts only known elements and also `ComponentType`.

jonkoops (changes requested):

packages/asc-ui/src/components/Header/HeaderLogoText.tsx
robinpiets: Done

jonkoops (approved):

jonkoops (approved):

PR #1415, opened 2020-12-02
Use correct type for as and forwardedAs props

jonkoops:


packages/asc-ui/src/components/Typography/TypographyStyle.ts
```diff
@@ -24,11 +25,16 @@ export const defaultTypographyStyles = {
...
 
-const getProperty = <T, K extends keyof T>(obj: T, key: K) => obj[key]
-
 export default styled.p<Props>`
-  ${({ as, forwardedAs }) =>
-    getProperty(defaultTypographyStyles, as || forwardedAs)};
+  ${({ as, forwardedAs }) => {
+    const key = as ?? forwardedAs
+
+    if (typeof key === 'string') {
```
robinpiets, line 32: Does this work?
`return typeof key === 'string' && defaultTypographyStyles[key] ?? null`?

jonkoops: Actually, the function can return any falsy value so we can even drop the `null`. Changed it.

packages/asc-ui/src/components/Typography/TypographyStyle.ts
```diff
@@ -7,28 +8,31 @@ export type Props = {
   element?: Variant
   fontSize?: number
   styleAs?: keyof Theme.TypographyElements
```
robinpiets, line 10: Can this be `ElementType` as well?

jonkoops: No we want to allow a specific subset of elements here that can only be used in typography,

robinpiets (approved):

PR #1453, opened 2020-12-23
Add support for React 17

jonkoops:


packages/asc-assets/package.json
```diff
@@ -26,6 +26,10 @@
     "build:react-icons": "svgr -d src/icons static/icons --template src/utils/generateIconReact.js --svg-props focusable={false} --ext tsx"
   },
   "sideEffects": false,
+  "peerDependencies": {
```
chrisvanmook, line 29: I don't think asc-assets is dependent on react, it can be used standalone without asc-ui. So this can be removed :)

jonkoops: It actually is dependent on React as the files generated include import statements to the `react` package.

However I am not quite sure that `react-dom` needs to be included as a peer dependency as no direct references are made both in `asc-ui` or `asc-assets`. I included it as a good measure, but it can be removed.

chrisvanmook (approved):

PR #1507, opened 2021-02-19
Changed how focus is set in modal

alimpens:
- set focus on first focussable element in modal
- reset focus to previously focussed element on modal close

- [ ] You've added or updated the README.md of the component
- [x] You have been following the guidelines, written in the [README](../blob/master/docs/CONTRIBUTING.md#user-content-conventions-and-rules) file
- [x] Your code has the necessary tests written
- [x] You've exported the component in [index.ts](./packages/asc-ui/src/index.ts)
- [x] You have updated the [CHANGELOG.md unreleased sections](../blob/master/CHANGELOG.md#muser-content-unreleased)


packages/asc-ui/src/components/Modal/Modal.tsx
```diff
@@ -36,28 +36,33 @@ const Modal: FunctionComponent<Props> = ({
...
-    }
-    return () => {
-      clearTimeout(renderedTimer)
+    if (open && ref.current) {
+      const node = ref.current
+      const firstFocusableEl = node.querySelector(FocusableElements.join(', '))
+
+      if (firstFocusableEl instanceof HTMLElement) {
+        firstFocusableEl.focus()
+      }
```
jonkoops, line 48: You can prevent some nesting here by using an early return.
```suggestion
    if (!open || !ref.current) {
      return
    }

    const node = ref.current
    const firstFocusableEl = node.querySelector(FocusableElements.join(', '))

    if (firstFocusableEl instanceof HTMLElement) {
      firstFocusableEl.focus()
    }
```

packages/asc-ui/src/components/Modal/Modal.tsx
```diff
@@ -36,28 +36,33 @@ const Modal: FunctionComponent<Props> = ({
...
-      clearTimeout(renderedTimer)
-      renderedTimer = setTimeout(() => {
-        node.focus()
-      })
-    }
-    return () => {
-      clearTimeout(renderedTimer)
+    if (open && ref.current) {
+      const node = ref.current
+      const firstFocusableEl = node.querySelector(FocusableElements.join(', '))
```
jonkoops, line 44: Since the `node` variable is no longer used in callbacks we can instead use `ref.current` directly:
```suggestion
      const firstFocusableEl = ref.current.querySelector(FocusableElements.join(', '))
```

packages/asc-ui/src/components/Modal/Modal.tsx
```diff
@@ -36,28 +36,33 @@ const Modal: FunctionComponent<Props> = ({
...
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [])
+  }, [open])
 
   const handleClose = () => {
     if (onClose) {
+      if (
+        previouslyFocusedElement &&
+        previouslyFocusedElement instanceof HTMLElement
+      ) {
```
jonkoops, line 57: Since `instanceof` will also check for `undefined` and other empty values there is no need to do an extra check.
```suggestion
 if (previouslyFocusedElement instanceof HTMLElement) {
```

packages/asc-ui/src/index.ts
```diff
@@ -152,7 +151,6 @@ export {
   FilterBox,
   FilterOption,
   FilterTag,
-  Focus,
```
jonkoops, line 155: This is a breaking change, can you add it to the changelog?

packages/asc-ui/src/utils/hooks/useTrappedFocus.ts
```diff
@@ -2,7 +2,7 @@ import { KeyboardEvent, RefObject } from 'react'
 import { KeyboardKeys } from '../../types'
 import ownerDocument from '../ownerDocument'
 
-const FocusableElements = [
+export const FocusableElements = [
```
jonkoops, line 5: Can you rename this one to use camel case?
```suggestion
export const focusableElements = [
```

packages/asc-ui/src/components/Modal/Modal.tsx
```diff
@@ -80,18 +85,17 @@ const Modal: FunctionComponent<Props> = ({
...
-          </ModalStyle>
-        </ModalStyleContainer>
-      </ModalFocus>
+      <BackDrop backdropOpacity={backdropOpacity} onClick={handleClose} />
+      <ModalStyle
+        role="dialog"
+        aria-modal="true"
+        ref={ref}
+        onKeyDown={handleKeyDown}
+        className={className}
```
jonkoops, line 94: Let's remove `className` from here and pass it with `otherProps` instead.

packages/asc-ui/src/components/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,69 @@
+import { fireEvent, render } from '@testing-library/react'
```
jonkoops, line 1: Thanks for adding these tests, very nice!

jonkoops (changes requested):

packages/asc-ui/src/index.ts
alimpens: Done!

packages/asc-ui/src/components/Modal/Modal.tsx
```diff
@@ -30,34 +31,39 @@ const Modal: FunctionComponent<Props> = ({
...
-  }, [])
+  }, [open])
 
   const handleClose = () => {
     if (onClose) {
+      if (previouslyFocusedElement instanceof HTMLElement) {
+        previouslyFocusedElement.focus()
+      }
       onClose()
     }
```
jonkoops, line 62: Can we add some tests to verify this re-focus logic is working correctly?

jonkoops (changes requested):
I'm only missing a test, but this PR looks good to me :+1:

packages/asc-ui/src/components/Modal/Modal.tsx
alimpens: Good call, I added a test

jonkoops (approved):

PR #1528, opened 2021-03-02
Export hooks directly from root of package

jasperswart:


jonkoops (changes requested):
This change looks good to me, @jasperswart could you add an entry to the changelog?

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ Prefix the change with one of these keywords:
 - _Fixed_: for any bug fixes.
 - _Security_: in case of vulnerabilities.
 
+## [new version]
+- Changed: Hooks should be exported directly [#1120]
```
chrisvanmook, line 18: I would suggest something like: "Breaking: `hooks` object is not exported anymore. You can import hooks now directly, for example `import { useFocus } from '@amsterdam/asc-ui' `"

jasperswart: gedaan

jonkoops (approved):

PR #1549, opened 2021-03-16
Fixed taskflow button bug for Safari 10

robinpiets:
I did some tests with the latest canary release in my own project. I found two bugs in Safari that I tried to solve in my previous PR #1548. Safari 10 had an issue with the line-height and Safari 14 still has same issue with the focus.

O/t: I think it's not easy to browser-test asc-ui in Storybook in the localhost setup. I would be really nice if Netlify builds each PR. But that's a different topic.. 

Safari 10 bug:
image

Safari 10 bug solved:
image

Safari 14 bug (still remains):
image

I tried a lot to solve the Safari 14 bug, but unfortunately it remains.


jonkoops (approved):

jonkoops:
LGTM!

PR #1552, opened 2021-03-18
added Toggle component

jasperswart:
- [ ] You've added or updated the README.mdx of the story of the component
- [x] You have been following the guidelines, written in the [README](../blob/master/docs/CONTRIBUTING.md#user-content-conventions-and-rules) file
- [x] Your code has the necessary tests written
- [x] You've exported the component in [index.ts](./packages/asc-ui/src/index.ts)
- [x] You have updated the [CHANGELOG.md unreleased sections](../blob/master/CHANGELOG.md#muser-content-unreleased)


packages/asc-ui/src/components/MenuToggle/Toggle.test.tsx
```diff
@@ -0,0 +1,64 @@
+import { render, fireEvent } from '@testing-library/react'
+import Toggle from './Toggle'
+import { KeyboardKeys } from '../../types'
+
+describe('Toggle', () => {
```
timvanoostrom, line 5: Can you rename Toggle to MenuToggle everywhere applicable? It's kind of confusing to me. Am I looking at a test foor Toggle or MenuToggle? Also, why is there a test for MenuToggle in the Toggle component PR?

packages/asc-ui/src/components/MenuToggle/Toggle.tsx
```diff
@@ -0,0 +1,160 @@
+import {
```
timvanoostrom, line 1: Did you add this component as well or did something go wrong with renaming the folder? It might be useful to use the `git mv <old name> <new name>` command to rename a folder / file

packages/asc-ui/src/components/Toggle/Toggle.tsx
```diff
@@ -1,160 +1,37 @@
...
-import ToggleStyle, { Props as ToggleStyleProps } from './ToggleStyle'
-import ownerDocument from '../../utils/ownerDocument'
-import usePassPropsToChildren from '../../utils/hooks/usePassPropsToChildren'
-import useActionOnEscape from '../../utils/hooks/useActionOnEscape'
-import ToggleButton, {
-  Props as ToggleButtonProps,
-} from '../Button/ToggleButton/ToggleButton'
-import BackDrop, { Props as BackDropProps } from '../BackDrop/BackDrop'
+import { forwardRef, HTMLAttributes, useImperativeHandle, useRef } from 'react'
+import RadioStyle, { Props, LabelStyle, WrapperStyle } from './ToggleStyle'
```
timvanoostrom, line 2: Is it a convention to separate these styles into another file? Because this component is so small I'd prefer placing all the components in 1 file.

packages/asc-ui/src/components/Toggle/Toggle.tsx
```diff
@@ -1,160 +1,37 @@
...
-      {render ? children : conditionalRenderedChildren}
-      {hasBackDrop && open && (
-        <BackDrop
-          onClick={handleOnClick}
-          zIndexOffset={zIndexOffset}
-          backdropOpacity={backdropOpacity}
-        />
-      )}
-    </ToggleStyle>
+    <WrapperStyle className={className}>
```
timvanoostrom, line 13: The MDN page about the Switch Role has an example of implementation like:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role

```
<button type="button" role="switch" aria-checked="true"
    id="speakerPower" class="switch">
    <span>off</span>
    <span>on</span>
</button>
<label for="speakerPower" class="switch">Speaker power</label>
```

It seems like you have implemented a component with 2 switches with the same name. I think it's preferable to implement it with the button component and use it like a boolean control.

Something like:

```
<ToggleButton onToggle={(event, state: 'on'|'off') => { ... }} />
```

timvanoostrom (changes requested):
@jasperswart, it's difficult to review the actual component you want to add here. The move/rename of the original Toggle folder messes up the readability of this PR.

Maybe just recreate the PR and don't rename the MenuToggle component and instead just name the Toggle to ToggleButton or something?

If you insist on renaming, which is fine by me, create a different PR for the rename and use the `git mv` command for proper renaming.

packages/asc-ui/src/components/MenuToggle/Toggle.test.tsx
jasperswart: yes I will do that

packages/asc-ui/src/components/MenuToggle/Toggle.tsx
jasperswart: sorry for that

packages/asc-ui/src/components/Toggle/Toggle.tsx
jasperswart: wouldn't it be better to keep de radio-buttons because it is syntactically better for the two state toggle buttons. that is what I did for the Switch component (checkbox).

packages/asc-ui/src/components/Toggle/Toggle.tsx
jasperswart: are you sure about that? does Jon agree about this as well?

packages/asc-ui/src/components/MenuToggle/Toggle.test.tsx
jasperswart: finished !

packages/asc-ui/src/components/Toggle/Toggle.tsx
jasperswart: maybe remove de role="switch" from the code

timvanoostrom: The Toggle is much like the switch. The only thing that's different is that you provide a fixed set of optional values for either state instead of using boolean state. 

My point being more that if you want to use`role="switch"`, the implementation should be different. Semantically it reads like:

Switch 1: value A [checked/unchecked] 
Switch 2: value B [checked/unchecked]

Instead of:

Switch: value [checked/unchecked]

packages/asc-ui/src/components/Toggle/Toggle.tsx
timvanoostrom: I don't know if he's the keeper of an convention. My preference is move them into 1 file. IMO It improves readability.

jasperswart: gedaan

packages/asc-ui/src/components/Toggle/Toggle.tsx
jasperswart: Yes but you can only have 1 value: you can never switch both switches to true.  the switches are not separated from each other. 

 That is why it should be a radio button instead. don't you agree?

timvanoostrom: No, I don't think radios are the best fit for this kind of switch/toggle/whatever-construct-with-2fixed-choices. You can remove the role="switch" if you want to continue with the radios. Radio's have their own aria semantics so no need for mixing those.

timvanoostrom:
Also, the rename of Toggle to MenuToggle is a breaking change. It's less intrusive to Not rename the Toggle button and just give another name to the newly created Toggle like I suggested:

`Maybe just recreate the PR and don't rename the MenuToggle component and instead just name the Toggle to ToggleButton or something?`

packages/asc-ui/src/components/Toggle/Toggle.tsx
jasperswart: I do not agree but I will do it anyway

timvanoostrom: Then you shouldn't. Maybe find a second opinion?

jasperswart: goed idee: ik vraag jon even

jonkoops: Yeah, I think I have to agree with @timvanoostrom here. We want to have a single element that is responsible for the `switch` role. Looking at examples on the web the most common way to do this seems to be to add the `switch` role to a `button` element, and setting `aria-checked` to the right value when it is pressed.

I think this makes the most sense as well, so I would suggest we follow the implementation as specified on [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role).

jasperswart: ok I will do that!

PR #1653, opened 2021-06-15
Feature/pagination

fortunato:
I've added a Pagination bar component as per the design system:
https://designsystem.amsterdam.nl/7awj1hc9f/p/431c4a-pagination

I still need to align what I've produced here with the design team.


packages/asc-ui/src/components/Pagination/PaginationStyle.ts
```diff
@@ -0,0 +1,75 @@
...
+`
+
+export const PageNumber = styled.button.attrs({
+  type: 'button',
+})<ButtonProps & React.HTMLAttributes<HTMLButtonElement>>`
+  cursor: pointer;
+  font-size: 16px;
+  font-weight: 400;
+  line-height: 22px;
+  padding: 0 10px;
```
chrisvanmook, line 45: I think we should use 12px here. Design system is made with 5px, but we made an agreement that components should be build on a unit of 4px. Therefore we could use `themeSpacing(3)` here

packages/asc-ui/src/components/Pagination/PaginationStyle.ts
```diff
@@ -0,0 +1,75 @@
...
+  ul {
+    display: flex;
+    align-items: center;
+    flex-wrap: wrap;
+    list-style: none;
+  }
+
+  li {
+    display: list-item;
+    margin: 0 15px 0 0;
```
chrisvanmook, line 19: I think we should use 16px here. Design system is made with 5px, but we made an agreement that components should be build on a unit of 4px. Therefore we could use themeSpacing(4) here

packages/asc-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,143 @@
...
+    )
+  }
+
+  return (
+    <PaginationStyle
+      collectionSize={collectionSize}
+      page={page}
+      pageSize={pageSize}
+      {...otherProps}
+      aria-label="paginatie navigatie"
```
chrisvanmook, line 101: This should also have a `role="navigation"` attribute: https://www.a11ymatters.com/pattern/pagination/

packages/asc-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,143 @@
...
+    )
+  }
+
+  const renderPhased = () => {
+    return (
+      <>
+        {currentPage > 1 && <li key="pag-1">{renderButton(1)}</li>}
+        {currentPage > 3 && (
+          <li key="pag-spacer-1" className="spacer">
+            ...
```
chrisvanmook, line 77: I think we should use the `&mldr;` ellipsis punctuation character here for semantic purposes

packages/asc-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,143 @@
...
+      </>
+    )
+  }
+
+  const renderPhased = () => {
+    return (
+      <>
+        {currentPage > 1 && <li key="pag-1">{renderButton(1)}</li>}
+        {currentPage > 3 && (
+          <li key="pag-spacer-1" className="spacer">
```
chrisvanmook, line 76: Why do we need a className here? If we need to target this in a test then we should rather use a testid

packages/asc-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,143 @@
...
+  }
+  if (totalPages <= 5 || currentPage < totalPages - 1) {
+    pageCluster.push(currentPage + 1)
+  }
+
+  React.useEffect(() => {
+    if (onPageChange !== undefined) onPageChange(currentPage)
+  }, [currentPage, onPageChange])
+
+  const handlePrevious = () => {
```
chrisvanmook, line 31: small thing, but could we rename this to `onHandlePrevious`? That would make it a bit more clear that this is an event handler

packages/asc-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,143 @@
...
+    setCurrentPage(currentPage - 1)
+  }
+
+  const handleNext = () => {
```
chrisvanmook, line 35: same here

packages/asc-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,143 @@
...
+    setCurrentPage(currentPage + 1)
+  }
+
+  const renderButton = (pageNumber: number) => {
```
chrisvanmook, line 39: We should never add a component inside of a component, as this would cause major performance issues. Best is to introduce a separate component (could be inside this file tho) called something like `PaginationButton` where it accepts a `pageNumber`

packages/asc-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,143 @@
...
+        }}
+        isCurrent={isCurrent}
+        tabIndex={isCurrent ? -1 : 0}
+      >
+        {pageNumber}
+      </PageNumber>
+    )
+  }
+
+  const renderUnphased = () => {
```
chrisvanmook, line 61: same here

packages/asc-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,143 @@
...
+    return (
+      <>
+        {Array.from(Array(totalPages).keys()).map((num) => (
+          <li key={`pag-${num}`}>{renderButton(num + 1)}</li>
+        ))}
+      </>
+    )
+  }
+
+  const renderPhased = () => {
```
chrisvanmook, line 71: and same here

packages/asc-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,143 @@
...
+    return (
+      <PageNumber
+        aria-label={
+          isCurrent ? `Pagina ${pageNumber}` : `Ga naar pagina ${pageNumber}`
+        }
+        aria-current={isCurrent}
+        data-testid={`btn-page-${pageNumber}`}
+        onClick={(evt) => {
+          // Remove focus from button after click
+          evt.currentTarget.blur()
```
chrisvanmook, line 50: why do we need to blur after click? I think that's unintended behaviour or am I missing something?

packages/asc-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,143 @@
...
+const Pagination: React.FC<Props & React.HTMLAttributes<HTMLElement>> = ({
+  page = 1,
+  collectionSize,
+  pageSize = 10,
+  onPageChange,
+  ...otherProps
+}) => {
+  const [currentPage, setCurrentPage] = React.useState<number>(page)
+  const totalPages = Math.ceil(collectionSize / pageSize)
+  const pageCluster = [currentPage]
```
chrisvanmook, line 19: Should this not be put inside a `useEffect`?

chrisvanmook (changes requested):

packages/asc-ui/src/components/Pagination/PaginationStyle.ts
fortunato: Fixed

fortunato: Fixed

packages/asc-ui/src/components/Pagination/Pagination.tsx
fortunato: Fixed

packages/asc-ui/src/components/Pagination/Pagination.tsx
fortunato: In this case I'm not sure if the 60 lines of additional code are a lesser evil than the (in this instance) small performance overhead. But I've refactored as you suggested

packages/asc-ui/src/components/Pagination/Pagination.tsx
fortunato: No longer necessary after getting rid of the yellow focused state

packages/asc-ui/src/components/Pagination/Pagination.tsx
fortunato: Check

packages/asc-ui/src/components/Pagination/Pagination.tsx
fortunato: Check

packages/asc-ui/src/components/Pagination/Pagination.tsx
fortunato: Nice catch; I thought I was going to need it but it is indeed unused. Removed.

packages/asc-ui/src/components/Pagination/Pagination.tsx
fortunato: Do you have a suggestion on how to work around this?
image

packages/asc-ui/src/components/Pagination/PaginationStyle.ts
fortunato: Fixed

packages/asc-ui/src/components/Pagination/Pagination.tsx
fortunato: Tbh I'm not 100% clear on when it's ok to refrain from using useEffect in such scenarios. The computational effort here seems negligible to me. What's a good rationale /rule of thumb here? Can you have a look at what I converted it to now?

packages/asc-ui/src/components/Pagination/Pagination.tsx
fortunato: I settled on `onPrevious`.

jonkoops (approved):

PR #1655, opened 2021-06-16
Change radio/checkbox components wrapper tag

MauritsR:
When using a `<Checkbox />` or `<Radio />` component, we have to wrap the component with a label to make sure that both the label and the input are clickable.

``` 
// Radio.stories.mdx
<RadioGroup name="group">
  <Label htmlFor="radio-1" label="Default">
    <Radio id="radio-1" />
  </Label>
  <Label htmlFor="radio-2" label="Checked">
    <Radio id="radio-2" checked />
  </Label>
</RadioGroup>
```

However, this means that labels contain a `div` tag (In `<Radio />`), which is invalid HTML: `div` is not part of the [content model](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) of a label. This PR changes it to a `span`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Prefix the change with one of these keywords:
 - Removed: Console warning in `ContextMenu` for not passing the right children
 - Removed: Support for React 16
 - Removed: `ViewerContainer` component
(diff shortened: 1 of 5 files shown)


timvanoostrom (approved):
LGTM

chrisvanmook (approved):

PR #1711, opened 2021-07-09
bugfix css Alert

remyvdwereld:


packages/asc-ui/src/components/Alert/AlertStyle.ts
```diff
@@ -41,6 +41,9 @@ export const ContentWrapper = styled.div`
   flex-direction: column;
   justify-content: center;
   width: 100%;
+  overflow-wrap: anywhere;
+  overflow: hidden;
+  margin-right: ${themeSpacing(4)};
```
chrisvanmook, line 46: I don't think we will need this margin, as it's will cause problems on smaller screens:
image
Versus current styling:
image

To be honest, I think we should refactor this component so the close button will float right and the test will wrap around the close button. I've tried to change the position of the close button in the HTML (next to the header / children), gave it a `float: right` and removed the styling from it's parent (that had `display: flex`). This way the paragraph can also use the full width!
image
I'm not sure if this will cause any other problems yet tho, but this could be a potential solution and would kill 2 birds with one stone.

remyvdwereld: Thanks @chrisvanmook for your feedback (love it!) and your patience! I had a week off. ;) I changed the PR. The `overflow-wrap: anywhere;` is needed for a long contiguous text but is not supported for Safari on IOS. I added the `overflow: hidden;` for these specific cases.

chrisvanmook (approved):

PR #1936, opened 2022-01-13
Setup beta

alimpens:

- [ ] You've added or updated the README.mdx of the story of the component
- [ ] You have been following the guidelines, written in the [README](../blob/main/docs/CONTRIBUTING.md#user-content-conventions-and-rules) file
- [ ] You've exported the component in [index.ts](./packages/asc-ui/src/index.ts)
- [ ] You have updated the [CHANGELOG.md unreleased sections](../blob/main/CHANGELOG.md#muser-content-unreleased)


package.json
```diff
@@ -6,11 +6,12 @@
   ],
   "scripts": {
     "prebuild": "npm run build --workspace=@amsterdam/asc-assets",
-    "build": "npm run clean:generated && npm run build:ts && npm run build:ts:es",
-    "build:watch": "npm run clean:generated && concurrently \"npm run build:ts -- --watch\" \"npm run build:ts:es -- --watch\"",
+    "build": "npm run clean:generated && npm run build:ts && npm run build:ts:es && npm run build:ts:beta",
+    "build:watch": "npm run clean:generated && concurrently \"npm run build:ts -- --watch\" \"npm run build:ts:es -- --watch\" \"npm run build:ts:beta -- --watch\"",
```
janjaap, line 10: Wouldn't it be better to split this into two separate scripts? Might be a lot to watch for if the `beta` folder grows and, when developing, I don't think you even need to watch for changes in both.

packages/asc-assets/static/fonts/beta-fonts.css
```diff
@@ -0,0 +1,57 @@
+@import url("https://fast.fonts.net/t/1.css?apiType=css&projectid=88ed95f9-d020-4dee-be83-444c3ada1a0e");
```
janjaap, line 1: Shouldn't this file (`beta-fonts.css`) also be placed in a beta folder? Like so: `packages/asc-assets/beta/static/fonts/beta-fonts.css`?

packages/asc-ui/src/beta/components/Button/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './Button'
```
janjaap, line 1: Shouldn't this export also contain the named exports?

janjaap (commented):
I think that any file, that is a duplicate from the original `asc-ui` package, should be placed in it's own shared package. This package shouldn't be published, but its contents should be accessible to the other packages.

packages/asc-assets/static/fonts/beta-fonts.css
alimpens: I dunno, does it have negative consequences to keep it where it is? We probably don't need anything else in the `beta/static` folder, so it would be a folder containing a single folder... Or maybe something like `packages/asc-assets/static/fonts/beta`?

packages/asc-ui/src/beta/components/Button/index.ts
alimpens: Do you mean the current export should be named? So `export { default as Button } from './Button'`? Or that the named exports in `ButtonStyle.ts` should be exported?

package.json
```diff
@@ -7,21 +7,21 @@
   "scripts": {
     "prebuild": "npm run build --workspace=@amsterdam/asc-assets",
     "build": "npm run clean:generated && npm run build:ts && npm run build:ts:es && npm run build:ts:beta",
-    "build:watch": "npm run clean:generated && concurrently \"npm run build:ts -- --watch\" \"npm run build:ts:es -- --watch\"",
-    "build:watch:beta": "npm run clean:generated && npm run build:ts:beta -- --watch",
+    "build:watch": "npm run clean:generated && concurrently \"npm run build:ts -- --watch\" \"npm run build:ts:es -- --watch\" \"npm run build:ts:beta\"",
+    "build:beta:watch": "npm run clean:generated && concurrently \"npm run build:ts\" \"npm run build:ts:es\" \"npm run build:ts:beta  -- --watch\"",
```
janjaap, line 11: Logical change, but not in-line with the new script commands below

packages/asc-ui-beta/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "homepage": "https://github.com/Amsterdam/amsterdam-styled-components#readme",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Amsterdam/amsterdam-styled-components.git",
+    "directory": "packages/asc-ui-beta"
+  },
+  "bugs": {
+    "url": "https://github.com/Amsterdam/amsterdam-styled-components/issues"
+  },
+  "version": "0.35.0",
```
janjaap, line 15: This version isn't correct, or is it? Do you want to keep it in sync with `asc-ui`?

alimpens: You're right, it should be `0.1.0`

package.json
alimpens: Do you mean the order of keywords? Maybe `build:ts:beta` should be `build:beta`, seeing as it build both `ts` and `es`

packages/asc-ui-beta/src/utils/__test__/betaUtils.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+  fluidTypoStyle,
+  typographyStyle,
+} from '../betaUtils'
+
+describe('calculateFluidStyle', () => {
+  it('should return the correct css clamp rule', () => {
+    const rule = calculateFluidStyle(16, 32)
+
+    expect(rule).toEqual(
+      `clamp(    1rem,    1rem + 1 *     (100vw - 20rem) / 100,    2rem  )`,
```
janjaap, line 16: Spacing looks a bit weird, but it's a test, so that shouldn't matter.

packages/asc-ui-beta/src/utils/betaUtils.ts
```diff
@@ -59,10 +59,7 @@ export const fluidTypoStyle = css`
 `
 
 export interface TypoStyle {
-  gutterBottom?: {
-    small: number
-    large: number
-  }
+  gutterBottom?: any
```
janjaap, line 62: Any particular reason for the prop `butterBottom` not to have a type anymore?

janjaap (approved):
A couple of remarks. Also, the coverage of the beta utils isn't a 100%; several branches and else paths aren't covered. Not a big issue, but something to keep an eye out for; wouldn't want to have too low a coverage.

packages/asc-ui-beta/src/utils/betaUtils.ts
alimpens: Yeah, the previous typing didn't work for both an object and a number :) I couldn't get the typing to work correctly, so I removed it for now. It's on my 'fix later' list

